PR #98, opened 2018-10-11
Adds favoriting to libraries and sorts libraries by favorites

KevWho:
Adds favoriting feature to Libraries table of AcademicViewController. Cells for libraries now have a heart on the right which users can tap to favorite libraries.
screen shot 2018-10-11 at 2 05 57 pm
Favorited libraries will automatically move to the top of the list.


berkeleyMobileiOS/Classes/Academics/AcademicsViewController/AcademicViewController.swift
```diff
@@ -71,8 +82,8 @@ class AcademicViewController: UIViewController, UITableViewDelegate, UITableView
                 }
                 
                 self.libraries = nonEmptyList as! [Library]
-                if let t = self.resourceTableView {
-                    t.reloadData()
+                if self.resourceTableView != nil {
```
kbunarjo: don't need the check here (`resourceTableView` is implicitly unwrapped) . if the outlet is broken, we'll just let the app crash.

PR #147, opened 2020-02-18
Create 'Filter' buttons to be used to filter halls, dining menus, etc.

KevWho:
- Creates FilterView, which displays a list of labels as a row of toggle-able buttons.
- Adds simple Filter struct with static method to apply multiple filters to a list of items.
- Includes SampleViewController, which demonstrates using the classes to filter a list of Ints (to remove after Dining / Library views completed).
Screen Shot 2020-02-17 at 7 33 26 PM

### DIFF
--- a/bm-persona/Assets/Colors.swift
+++ b/bm-persona/Assets/Colors.swift
@@ -63,6 +63,10 @@ struct Color {
         return UIColor(displayP3Red: 255/255, green: 255/255, blue: 255/255, alpha: 1.0)
     }
     
+    static var selectedButtonBackground: UIColor {
+        return UIColor(displayP3Red: 251/255, green: 155/255, blue: 142/255, alpha: 1.0)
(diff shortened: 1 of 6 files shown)


bm-persona/Assets/Colors.swift
```diff
@@ -63,6 +63,10 @@ struct Color {
         return UIColor(displayP3Red: 255/255, green: 255/255, blue: 255/255, alpha: 1.0)
     }
     
+    static var selectedButtonBackground: UIColor {
+        return UIColor(displayP3Red: 251/255, green: 155/255, blue: 142/255, alpha: 1.0)
```
Stereopoly: At some point we should probably add all the dark theme colors

KevWho: You're right, but I don't know what that looks like yet :(

PR #148, opened 2020-02-25
Oscar/calendar

Stereopoly:


bm-persona/Common/EventTableViewCell.swift
```diff
@@ -0,0 +1,186 @@
...
+    override func setSelected(_ selected: Bool, animated: Bool) {
+        super.setSelected(selected, animated: animated)
+
+        // Configure the view for the selected state
+    }
+
+}
+
+
+extension UILabel {
```
KevWho: May want to move this to a new Utils file. Also, may want to look into Common/TagView, which does something similar.

PR #149, opened 2020-03-06
Map Marker Icons

KevWho:
Adds Markers to Map.
- Adds `MapMarker` model class
- Adds `MapDataSource` file that fetches an array of all `MapMarkers`
- Adds marker icons and slightly reorganizes Assets.xcassets
- Adds a `FilterView` to `MapViewController` to allow users to select a type of marker
- When a marker is selected, populates and displays a `MapMarkerDetailView` and hides the drawer through `DrawerContainer` protocol
- Reorganizes some code.

Screen Shot 2020-03-06 at 12 33 31 AM
Screen Shot 2020-03-06 at 12 33 47 AM

Future improvements:
- Update `MapMarkerDetail.openNow` with correct icon (currently "Placemark", should be clock icon)
- Add transit info
- Add distance estimates
- Change annotation after tapping marker
- Improve efficiency of loading / selecting markers and populating detail view


bm-persona/Common/TagView.swift
```diff
@@ -51,3 +51,19 @@ class TagView: UILabel {
     }
 
 }
+
+// Common TagViews
+extension TagView {
+    
+    static var open: TagView {
+        let color = UIColor(displayP3Red: 133/255, green: 158/255, blue: 248/255, alpha: 1.0)
```
Stereopoly: Move these colors to Assets/Color.swift?

bm-persona/Map/MapDataSource/MapMarker.swift
```diff
@@ -0,0 +1,117 @@
...
+            break
+        case .bikes:
+            icon = UIImage(named: "bike-icon")
+            break
+        }
+        return (icon ?? UIImage()).resized(size: CGSize(width: 20, height: 20))
+    }
+    
+    /** The color describing this marker type */
+    func color() -> UIColor {
```
Stereopoly: We should ask design for updated colors that match the Persona redesign

bm-persona/Common/TagView.swift
KevWho: Done!

bm-persona/Map/MapViewController.swift
```diff
@@ -102,9 +123,79 @@ class MapViewController: UIViewController {
...
+    }
+    
 }
 
+
+// MARK: - CLLocationManagerDelegate
+
 extension MapViewController: CLLocationManagerDelegate {
     func locationManager(_ manager: CLLocationManager, didChangeAuthorization status: CLAuthorizationStatus) {
         if status == .authorizedWhenInUse {
```
Stereopoly: Add cases if user declines perms etc
[Auth status docs](https://developer.apple.com/documentation/corelocation/clauthorizationstatus)

PR #151, opened 2020-03-13
Oscar/resources

Stereopoly:
Simulator Screen Shot - iPhone 11 Pro Max - 2020-03-12 at 19 01 12

- Resources page with data fetched from Firebase


Stereopoly:
Backend is missing most data needed in the design.

sidthesquid:
Nit: maybe set the inner card shadows to be basically the same as the outer container?

Stereopoly:
Shadow now matches card. Maybe we should have some universal constants or smt so all the shadows are the same? @KevWho

sidthesquid:
Yeah, I was thinking it would be similar to Shawn's PR. Should definitely have some universal views and constants. For example, the border radius of the cards, shadow, padding, etc. should be same across screens.

bm-persona/Map/MapViewController.swift
```diff
@@ -142,7 +142,7 @@ extension MapViewController: SearchBarDelegate {
         guard !searchResultsView.isScrolling else { return }
         showSearchResultsView(false)
         searchBar.setButtonStates(hasInput: textField.text?.count != 0, isSearching: false)
-    }
+    }
```
KevWho: 🔥

bm-persona/Common/ResourceTableViewCell.swift
```diff
@@ -0,0 +1,107 @@
...
+        get {
+            return super.frame
+        }
+        set (newFrame) {
+            var frame = newFrame
+            frame.origin.x += 5
+            frame.size.width -= 10
+            super.frame = frame
+        }
+    }
```
KevWho: Same as I mentioned in Shawn's PR - may want to consider just using frame.inset

PR #156, opened 2020-03-17
Hotfix - DataManager bad access

KevWho:
- Adds `AtomicDictionary`, a simple wrapper for a dictionary with limited functionality that employs a read-write lock to prevent bad accesses due to concurrency.
- Changes dictionaries in `MapDataSource` and in `DataManager` to `AtomicDictionaries`.


bm-persona/Data/DataManager.swift
```diff
@@ -8,6 +8,7 @@
 
 import Foundation
 
+/** The list of `DataSource` classes to fetch data from. Add to this list to add new data to the app. */
```
Stereopoly: Style: Using /** or /* ?

PR #160, opened 2020-04-03
DataSource Changes and Additions

KevWho:
- Fix Firebase collection names for Map Markers
- Add `GymClassDataSource`
    -  Currently gets classes in the week of 2019-02-24 if running in debug mode, since there is no up-to-date data on Firebase at the time of this PR.
- Add Dining models and `DiningHallDataSource` (still need to add cafes)
    - `DiningHallDataSource` currently gets menu data from 2020-03-14 if running in debug mode, since there is no up-to-date data on Firebase at the time of this PR.
- Remove `isOpen` for classes, add `HasOpenTimes` protocol and `WeeklyHours` class to unify opening times model for items


bm-persona/Resources/ResourcesDataSource/Resource.swift
```diff
@@ -32,25 +33,12 @@ class Resource: SearchItem, HasLocation, HasOpenTimes {
     var latitude: Double?
     var longitude: Double?
     
-    var isOpen: Bool {
```
Stereopoly, line 35: I like this clean up

bm-persona/Dining/DiningDataSource/DiningHalls/DiningHallDataSource.swift
```diff
@@ -0,0 +1,97 @@
...
+            requests.leave()
+        }
+        requests.notify(queue: .global(qos: .utility)) {
+            fetchMenus(diningHalls, completion: completion)
+        }
+    }
+    
+    static func fetchMenus(_ diningHalls: [String: DiningHall], completion: @escaping DataSource.completionHandler) {
+#if DEBUG
+        let collectionToday = "2020-03-14"
```
Stereopoly, line 47: Last day with dining hall info

bm-persona/Dining/DiningDataSource/DiningHalls/DiningHall.swift
```diff
@@ -0,0 +1,30 @@
...
+import Foundation
+
+/**
+    `DiningHall` represents a single physical dining location.
+    Each hall contains the `DiningMenu`, Open & Close times for every `MealType`.
+*/
+#warning("TODO: Make DiningHall implement SearchItem and HasLocation")
+class DiningHall: HasOpenTimes {
+    
+    let name: String
```
Stereopoly, line 18: How are we planning to track whether a place is favorited?

bm-persona/Dining/DiningDataSource/DiningHalls/DiningHallDataSource.swift
```diff
@@ -0,0 +1,97 @@
...
+import Firebase
+
+fileprivate let kDiningHallEndpoint = "Dining Halls"
+
+class DiningHallDataSource: DataSource {
+    
+    // Returns the collection name for the given date. Must match the name on Firebase.
+    private static func collectionForDate(_ date: Date) -> String {
+        let dateFormatter = DateFormatter()
+        dateFormatter.dateFormat = "yyyy-MM-dd"
```
Stereopoly, line 19: We should standardize which dateFormat we use. I think I was using MM/dd/yyyy

KevWho: I did this format because I use it to get the collection for the day's menus from Firebase, which has this format. Unsure if there's a better way to query the proper collection.

bm-persona/Dining/DiningDataSource/DiningHalls/DiningHall.swift
KevWho: Probably another protocol. The old codebase used a `Favorable` protocol (required the boolean `isFavorited`) with a `FavoriteStore` class that managed saving / reading the object's state from the device.

bm-persona/Dining/DiningDataSource/DiningItem.swift
```diff
@@ -0,0 +1,27 @@
...
+/** DiningMenu is a typealias for an array of DiningItems. */
+typealias DiningMenu = Array<DiningItem>
+
+/** Object representing a single dish/item. */
+class DiningItem {
+    
+    let name: String
+    
+    // Additional information regarding food such as allergens or alternative options
+    let restrictions: [String]
```
huangshawn1, line 20: do we want to save restrictions as strings? zeplin shows filters for gluten-free, vegan, etc. so we might want to use an enum for all the categories a dining item fits

bm-persona/Dining/DiningDataSource/DiningHalls/DiningHall.swift
Stereopoly: We can do this after initial release then

bm-persona/Dining/DiningDataSource/DiningItem.swift
Stereopoly: It's stored in Firebase as an array of strings so likely just placed into this.

KevWho: Yea. I just wanted to make sure we could still display new restrictions that the app isn't explicitly programmed to handle. I think the enum is a good idea, though, so I pushed a commit that adds a KnownRestriction enum (so we can display things like icons for the ones we know) and a DiningRestriction class that will allow us to handle both known and unknown restrictions.

PR #164, opened 2020-04-06
Shawn/dining detail

huangshawn1:
Simulator Screen Shot - iPhone 11 Pro Max - 2020-04-05 at 16 57 05
Simulator Screen Shot - iPhone 11 Pro Max - 2020-04-05 at 16 58 15
Created dining detail page with overview card and menu items.
Todo:
- Use actual capacities
- Save favoriting


bm-persona/Dining/DiningDetailViewController.swift
```diff
@@ -0,0 +1,496 @@
...
+
+import UIKit
+import CoreLocation
+
+fileprivate let kHeaderFont: UIFont = Font.bold(24)
+fileprivate let kCardPadding: UIEdgeInsets = UIEdgeInsets(top: 16, left: 16, bottom: 16, right: 16)
+fileprivate let kViewMargin: CGFloat = 16
+class DiningDetailViewController: UIViewController {
+    
+    var diningHall: DiningHall?
```
KevWho, line 17: What's the reason for making this an optional vs. an implicitly unwrapped optional? It seems like you're force-unwrapping `diningHall` whenever you use it, so there doesn't seem to be a purpose for making this an optional.

bm-persona/Dining/DiningDetailViewController.swift
```diff
@@ -0,0 +1,496 @@
...
+//  bm-persona
+//
+//  Created by Shawn Huang on 4/4/20.
+//  Copyright © 2020 RJ Pimentel. All rights reserved.
+//
+
+import UIKit
+import CoreLocation
+
+fileprivate let kHeaderFont: UIFont = Font.bold(24)
```
KevWho, line 12: Doesn't appear to be used.

bm-persona/Dining/DiningDetailViewController.swift
```diff
@@ -0,0 +1,496 @@
...
+                        nextOpenInterval = interval
+                    } else if interval.compare(nextOpenInterval!) == .orderedAscending {
+                        nextOpenInterval = interval
+                    }
+                }
+            }
+            if nextOpenInterval != nil {
+                openTimeLabel.text = formatter.string(from: nextOpenInterval!.start, to: nextOpenInterval!.end)
+            }
+        }
```
KevWho, line 237: The intervals should be in sorted order, but this works too, so I guess it's fine.

bm-persona/Dining/DiningDetailViewController.swift
```diff
@@ -0,0 +1,496 @@
...
+        if diningHall!.isOpen! {
+            openTag.text = "Open"
+            openTag.backgroundColor = Color.openTag
+        } else {
+            openTag.text = "Closed"
+            openTag.backgroundColor = Color.closedTag
+        }
+        openTag.leftAnchor.constraint(equalTo: openTimeLabel.rightAnchor, constant: kViewMargin).isActive = true
+        openTag.centerYAnchor.constraint(equalTo: clockIcon.centerYAnchor).isActive = true
+        openTag.widthAnchor.constraint(equalToConstant: 50).isActive = true
```
KevWho, line 250: I guess for future reference, I'm pretty sure you don't need to specify a width constraint for a TagView unless you wanna.

bm-persona/Dining/DiningDetailViewController.swift
```diff
@@ -0,0 +1,496 @@
...
+                DispatchQueue.main.async {
+                    self.diningHall!.image = image
+                    self.diningImage.image = image
+                }
+            }
+        }
+        diningImage.rightAnchor.constraint(equalTo: card.layoutMarginsGuide.rightAnchor).isActive = true
+        diningImage.topAnchor.constraint(equalTo: card.layoutMarginsGuide.topAnchor).isActive = true
+        diningImage.bottomAnchor.constraint(equalTo: faveButton.topAnchor, constant: -1 * kViewMargin).isActive = true
+        diningImage.widthAnchor.constraint(equalTo: card.layoutMarginsGuide.widthAnchor, multiplier: 0.45).isActive = true
```
KevWho, line 291: Would it be better to constrain the aspect ratio of the image? Unsure, so lmk.

bm-persona/Dining/DiningDetailViewController.swift
huangshawn1: yeah this was because without the width constraint the tags would be too small or disappear if the opening times label is too long. I wanted to constrain everything so that the only thing that could change in size was the label

bm-persona/Dining/DiningDetailViewController.swift
huangshawn1: I think this is how you constrain the aspect ratio using anchors

KevWho: I mean the aspect ratio of the image width to the image height.

huangshawn1: oh yeah that's better

bm-persona/Dining/DiningDetailViewController.swift
```diff
@@ -0,0 +1,504 @@
...
+        setupMenu()
+        
+        if CLLocationManager.locationServicesEnabled() {
+            locationManager.requestLocation()
+        }
+    }
+    
+    let nameLabel: UILabel = {
+        let label = UILabel()
+        label.font = Font.semibold(24)
```
KevWho, line 44: Ah, sorry for forgetting to mention this, but could you make this label bold instead of semibold?

PR #165, opened 2020-04-06
Gym Classes

KevWho:
- Populate Gym Class section in Fitness View Controller (TableView and horizontal `CardCollectionView`)
- Split up Fitness View Controller into separate files, since the original file was getting bloated: `GymsController` and `GymClassesController` implement DataSource protocols for CollectionViews and TableViews for `FitnessViewController`.
- For the "Today's Classes" section, in Debug mode, picks some arbitrary day's data, since there is currently no up-to-date data in Firebase.
- Fix start and end times for gym classes
- Add `GymClassType` and colors for the class types (Cardio, Aqua, etc.)
- Change some constraints for `EventTableViewCell` and `CardCollectionViewCell`
image


bm-persona/Utils/Date+Extension.swift
```diff
@@ -6,11 +6,11 @@ import Foundation
  */
 extension Date
 {
-    // Returns today with (hour:min) of this Date. 
-    func sameTimeToday() -> Date?
+    // Returns `date` with (hour:min) of this Date.
+    func sameTime(on date: Date) -> Date?
     {
```
Stereopoly, line 11: Minor nit - braces on newline

PR #166, opened 2020-04-10
Hotfix: Gym Class selection

KevWho:
Previously, gym classes in the 'Today' tableview were unable to be deselected.
- Disable selection on 'Today' gym classes


bm-persona/Fitness/FitnessViewController.swift
```diff
@@ -210,6 +210,7 @@ extension FitnessViewController {
         scheduleButton.rightAnchor.constraint(equalTo: card.layoutMarginsGuide.rightAnchor).isActive = true
         
         classesTable = UITableView()
+        classesTable.allowsSelection = false
```
Stereopoly, line 213: Works as expected :)

PR #170, opened 2020-04-18
Hotfix: Library Cell Selection

KevWho:
- Re-enables selection for cells on LibraryViewController


bm-persona/Libraries/LibraryViewController.swift
```diff
@@ -120,7 +119,7 @@ class LibraryViewController: UIViewController, UITableViewDataSource, UITableVie
         let vc = LibraryDetailViewController()
         vc.library = self.filterTableView.filteredData[indexPath.row]
         present(vc, animated: true, completion: nil)
-
+        tableView.deselectRow(at: indexPath, animated: true)
```
Stereopoly, line 122: 💯

PR #171, opened 2020-04-19
Shawn/map search

huangshawn1:
Drop pin on map and show detail view when item is searched.
- SearchAnnotation: map annotation for search items
- MapViewController extension for SearchResultsViewDelegate: drops pin and initializes corresponding drawer

Ended up reworking how all the drawers work and changed the library and dining detail views to all use drawers now. Should be easier to implement drawers in the future.
- DrawerViewDelegate: Handles panning and moving drawers.
- MainDrawerViewDelegate: Handles a stack of all active drawers, hiding and showing them.
- SearchDrawerViewDelegate: Presents a detail view for a search item.

Also a slight fix for dining detail open times label for when the interval spans more than one day
Simulator Screen Shot - iPhone 11 Pro Max - 2020-04-18 at 22 14 43
Simulator Screen Shot - iPhone 11 Pro Max - 2020-04-18 at 22 14 54
Simulator Screen Shot - iPhone 11 Pro Max - 2020-04-18 at 22 15 03

Concerns/Todo:
- unify detail views to look more similar
- add collapsible open time card to library detail (so open time can be shown when detail view is in middle position)
- middle position covers a large portion of the map on smaller phones (like iPhone 8), there’s no way to dismiss the detail view without removing the pin and there's no lower position than middle
Simulator Screen Shot - iPhone 8 - 2020-04-18 at 22 01 14


bm-persona/Map/MapViewController.swift
```diff
@@ -167,19 +198,44 @@ extension MapViewController: MKMapViewDelegate {
...
     }
     
     func mapView(_ mapView: MKMapView, didSelect view: MKAnnotationView) {
-        markerDetail.marker = view.annotation as? MapMarker
-        drawerContainer?.moveDrawer(to: .hidden, duration: 0.2)
+        if let annotation = view.annotation as? MapMarker {
+            markerDetail.marker = annotation
+            // hide either search detail or drawer
+            if detailViewController != nil {
+                detailViewController!.view.isHidden = true
```
KevWho, line 219: Small nitpick, but when hiding / showing `detailViewController`, can we animate it down / up like the main drawer?

bm-persona/Dining/DiningDetailViewController.swift
```diff
@@ -33,6 +33,10 @@ class DiningDetailViewController: UIViewController {
         setUpOverviewCard()
         setupMenuControl()
         setupMenu()
+        view.layoutSubviews()
+        // set the bottom cutoff point for when view appears due to map search
+        // the "middle" position for the view will show everything in the overview card
+        middleCutoffPosition = card.frame.maxY + 5
```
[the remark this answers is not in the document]
huangshawn1: tested it out, the cutoff is used in MapViewController before viewDidLayoutSubviews is called

bm-persona/Map/MapViewController.swift
```diff
@@ -167,19 +198,44 @@ extension MapViewController: MKMapViewDelegate {
...
     
     func mapView(_ mapView: MKMapView, didDeselect view: MKAnnotationView) {
-        markerDetail.marker = nil
-        drawerContainer?.moveDrawer(to: .collapsed, duration: 0.2)
+        if (view.annotation as? MapMarker) != nil {
+            markerDetail.marker = nil
+            // show either search detail or drawer
+            if detailViewController != nil {
+                detailViewController!.view.isHidden = false
+                mapView.selectAnnotation(searchAnnotation!, animated: true)
```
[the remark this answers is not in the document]
huangshawn1: strange problem with the obvious fix you suggested: map marker is deselected, which selects the searchAnnotation, but then immediately the second map marker is selected. however, only one annotation can be selected at a time. what happens is after searchAnnotation is selected, the second map marker being selected somehow removes searchAnnotation from the array of selected annotations without deselecting it. the searchAnnotation grows big, but calling deselect on it doesn't do anything because it's no longer in the array of selected annotations. my fix was asynchronously checking if markerDetail.marker is nil before showing the detail view. not sure if there's a better way to do it.

bm-persona/Map/MapViewController.swift
huangshawn1: yeah I tried to do that before and tried again, but there's a problem and idk how to fix it. basically, calling moveDrawer(.hidden) moves it to hidden but something automatically moves it back to the middle position. asynchronously calling moveDrawer(.hidden) works but when selecting a second mapMarker, you can see the view flash up then down, so it looks like the view is moving up for some reason. I set a breakpoint in moveDrawer, but it only gets called once so I've got no clue why the view moves twice

huangshawn1: so there were some other problems with the movedrawer function, but essentially I tried moving the center of the drawer twice, once initially and once in the completion handler. somehow, this fixed this bug and some others. not sure why

bm-persona/Drawer/DrawerViewController.swift
```diff
@@ -8,55 +8,40 @@
...
+        
+        setupBackgroundView()
+        setupGestures()
     }
     
     func setupBackgroundView() {
-        let backgroundView = UIView()
-        backgroundView.backgroundColor = Color.modalBackground
+        view.backgroundColor = Color.modalBackground
+        view.layer.cornerRadius = 50
```
Stereopoly, line 35: Updated this to be 15. I think with that could we move the content in the drawer up a bit?

huangshawn1: updated and the content was so far down because of a constraint issue which is fixed. distance from the top is now much smaller

PR #177, opened 2020-06-07
Shawn/occupancy

huangshawn1:
Integrates occupancy data for libraries, dining halls, and gyms.
- OccupancyDataSource: fetches occupancy data and sets occupancy attribute for all available libraries, dining halls, and gyms
- Occupancy: matches day of week and hour of day to occupancy percent/status
- HasOccupancy: protocol for all objects that can have an occupancy percent
- Library, dining, and gym view controllers now use correct occupancies in the table. Dining detail view also uses correct occupancy. Waiting on library detail view's occupancy chart design to be finalized.

Other fixes.
- Changed DataManager fetch function and changed DataSource protocol to make sure all data sources only access Firebase once (even with concurrent calls to fetch). Pauses all other requests until the first one goes through.
- Fixed issue with dining detail view's distance icon/label being hidden.

Simulator Screen Shot - iPhone 11 Pro Max - 2020-06-07 at 16 13 40
Simulator Screen Shot - iPhone 11 Pro Max - 2020-06-07 at 16 14 00


bm-persona/Assets/Occupancy/Occupancy.swift
```diff
@@ -0,0 +1,57 @@
+//
+//  Occupancy.swift
```
Stereopoly, line 2: I don't think this should be under /Assets/ right

bm-persona/Dining/DiningDetailViewController.swift
```diff
@@ -265,9 +265,21 @@ extension DiningDetailViewController {
...
+                capBadge.backgroundColor = Color.highCapacityTag
+            case OccupancyStatus.medium:
+                capBadge.text = "Medium"
+                capBadge.backgroundColor = Color.medCapacityTag
+            case OccupancyStatus.low:
+                capBadge.text = "Low"
+                capBadge.backgroundColor = Color.lowCapacityTag
+            }
+        } else {
+            capBadge.isHidden = true
```
Stereopoly, line 281: Should we hide the dining table icon as well?

bm-persona/Dining/DiningViewController.swift
```diff
@@ -86,24 +97,22 @@ extension DiningViewController: UITableViewDelegate, UITableViewDataSource {
...
+                    cell.capBadge.backgroundColor = Color.highCapacityTag
+                case OccupancyStatus.medium:
+                    cell.capBadge.text = "Medium"
+                    cell.capBadge.backgroundColor = Color.medCapacityTag
+                case OccupancyStatus.low:
+                    cell.capBadge.text = "Low"
+                    cell.capBadge.backgroundColor = Color.lowCapacityTag
+                }
+            } else {
+                cell.capBadge.isHidden = true
```
Stereopoly, line 115: Hide the logo?

bm-persona/Assets/Occupancy/Occupancy.swift
huangshawn1: yeah, any thoughts on moving it to data vs having it in a separate folder

bm-persona/Dining/DiningDetailViewController.swift
huangshawn1: thought about that as well, we should probably establish what we're doing when stuff is unavailable in general. do we want to hide it, say n/a or unknown, rearrange other elements, etc. should be uniform throughout the app

huangshawn1:
> Also this update seems to have broken the gym page distances, as none of them show up for me anymore.

Gym distances don't show up for me on develop, could be to do with Firebase? Don't think I changed anything important relating to gym page but could be wrong.

bm-persona/Assets/Occupancy/Occupancy.swift
Stereopoly: Data

bm-persona/Dining/DiningDetailViewController.swift
Stereopoly: Yeah I guess for now we can just leave it

PR #178, opened 2020-06-08
Oscar/data enhancements

Stereopoly:
* Notice on events page when none exist

Simulator Screen Shot - iPhone 11 Pro - 2020-06-08 at 10 31 10

### DIFF
--- a/bm-persona/Common/FilterTableView/FilterTableView.swift
+++ b/bm-persona/Common/FilterTableView/FilterTableView.swift
@@ -101,23 +101,6 @@ class FilterTableView<T>: UIView {
     
 }
 
-extension FilterTableView {
-    func setupMissingView(parentView: UIView) {
-        let view = MissingDataView()
-        
-        view.isHidden = true
-        parentView.addSubview(view)
-        
-        view.translatesAutoresizingMaskIntoConstraints = false
-        view.topAnchor.constraint(equalTo: parentView.layoutMarginsGuide.topAnchor).isActive = true
-        view.leftAnchor.constraint(equalTo: parentView.layoutMarginsGuide.leftAnchor).isActive = true
-        view.rightAnchor.constraint(equalTo: parentView.layoutMarginsGuide.rightAnchor).isActive = true
-        view.bottomAnchor.constraint(equalTo: parentView.layoutMarginsGuide.bottomAnchor).isActive = true
-        
-        missingView = view
-    }
-}
-
 
 extension FilterTableView: FilterViewDelegate {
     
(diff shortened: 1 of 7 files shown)


bm-persona/Common/FilterTableView/FilterTableView.swift
```diff
@@ -87,6 +99,24 @@ class FilterTableView<T>: UIView {
...
+        
+        view.translatesAutoresizingMaskIntoConstraints = false
+        view.topAnchor.constraint(equalTo: self.layoutMarginsGuide.topAnchor).isActive = true
+        view.leftAnchor.constraint(equalTo: self.layoutMarginsGuide.leftAnchor).isActive = true
+        view.rightAnchor.constraint(equalTo: self.layoutMarginsGuide.rightAnchor).isActive = true
+        view.bottomAnchor.constraint(equalTo: self.layoutMarginsGuide.bottomAnchor).isActive = true
+        
+        missingView = view
+    }
+}
```
huangshawn1, line 117: this code is almost identical to what's happening for the calendar. maybe make this a function in MissingDataView class that takes in a UIView to be added to (calendarCard for CalendarVC and self for FilterTableView)

bm-persona/Common/MissingDataView.swift
```diff
@@ -0,0 +1,40 @@
...
+        label.text = "No items found"
+        
+        label.translatesAutoresizingMaskIntoConstraints = false
+        
+        label.center = self.center
+        
+        self.addSubview(label)
+        
+        label.centerXAnchor.constraint(equalTo: self.centerXAnchor).isActive = true
+        label.centerYAnchor.constraint(equalTo: self.centerYAnchor).isActive = true
```
huangshawn1, line 31: any reason to set the center and also constrain?

bm-persona/Common/MissingDataView.swift
```diff
@@ -0,0 +1,40 @@
...
+    private var missingLabel: UILabel!
+
+    override init(frame: CGRect) {
+        super.init(frame: frame)
+        
+        let label = UILabel()
+        label.textColor = Color.darkGrayText
+        label.textAlignment = .center
+        label.font = Font.medium(18)
+        label.text = "No items found"
```
huangshawn1, line 22: we should probably have a file to store all constant string messages like this one (for consistency in the future). not too relevant for this pr

huangshawn1: for example "Nothing found!" when search has no results or "Unknown" when there's no phone number

bm-persona/Common/FilterTableView/FilterTableView.swift
Stereopoly: Ah I remember what was happening - adding this view didn't work for some reason when using it in the FilterTableView

PR #181, opened 2020-07-01
dining detail crash hotfix + other fixes

huangshawn1:
Make menu control optional and check for non-nil value before using it.

Also fixed issues with MissingDataView:
- moved isHidden inside DispatchQueue to prevent error
- changed setup to use self instead of a newly instantiated MissingDataView. before, there was no way to hide the missing data view because the setup function added a new one to the parent view instead of adding itself.

Other changes:
- removed cafes from dining table because we're moving them to map markers
- changed cafe endpoint from "Cafes" to "Cafe" to match Firebase
- called filterTableView.update() instead of just reloading the table data: this makes sure that the missing data view is updated when data changes/filters are applied


Stereopoly:
The items aren't clickable anymore?

huangshawn1:
> The items aren't clickable anymore?

yeah because the missing data view is on top. if you comment that out you can test the change

PR #182, opened 2020-07-09
Open Times Card

huangshawn1:
Added collapsible card view (general view that can open and close) and open times card:

CollapsibleCardView
- collapsedView and openedView indicating what to display when opened and closed
- currently, tapping anywhere on the card will toggle the state. this can be changed to a specific region if we want (e.g. only in the top/collapsed part)

OpenTimesCardView
- displays the open times for any HasOpenTimes object using a CollapsibleCardView
- vertically stacks the days of the week and any intervals within the same day are also vertically stacked
- added some dummy data for testing in the LibraryDataSource. couldn't find any old library data on Firebase, so I just manually copied in some open close times from cafes/dining halls

Other changes:
- library detail cutoff for 'middle' drawer state is right under the open times card now
  - opening the open times card while in the 'middle' position will set the drawer to 'full'
- setting the drawer to 'full' will be done with 0.6 sec duration instead of 0.2 (by default, regardless of where/how the drawer is set to 'full' position). this is because the animation looked too fast at 0.2
- moved "next open interval" logic into HasOpenTimes because it's used in multiple places
- DetailView folder inside Common to hold all the views that are used for details
- got rid of logic that set place to be always open if the day contains any interval with duration 0
- changed how the scroll view in LibraryDetailVC works because it didn't work when I was working on the occupancy graphs. also cleaned up some other stuff in LibraryDetailVC.

Images (in order), slightly outdated - the book button width now matches the cards:
- closed all days (0 sec intervals)
- multiple intervals per day (stacked), bolded current interval, sunday has no data so left blank
- current time isn't in any interval for the day (no bold), friday closed all day (0 sec interval)
- collapsed

Simulator Screen Shot - iPhone 11 Pro Max - 2020-07-08 at 19 21 00
Simulator Screen Shot - iPhone 11 Pro Max - 2020-07-08 at 19 21 20
Simulator Screen Shot - iPhone 11 Pro Max - 2020-07-08 at 19 21 32
Simulator Screen Shot - iPhone 11 Pro Max - 2020-07-08 at 19 44 11


bm-persona/Common/DetailView/OpenTimesCardView.swift
```diff
@@ -0,0 +1,196 @@
...
+        var rightView: UIView?
+        
+        if let isOpen = item.isOpen {
+            if isOpen {
+                openTag.text = "Open"
+                openTag.backgroundColor = Color.openTag
+            } else {
+                openTag.text = "Closed"
+                openTag.backgroundColor = Color.closedTag
+            }
```
KevWho, line 43: There's also `TagView.open` and `TagView.closed` which has the text and colors you used here, so we could reduce this to: 
`leftView = isOpen ? TagView.open : TagView.closed`

(I think)

bm-persona/Common/DetailView/OpenTimesCardView.swift
```diff
@@ -0,0 +1,196 @@
...
+        guard leftView != nil || rightView != nil else { return nil }
+        let view = UIView()
+        view.translatesAutoresizingMaskIntoConstraints = false
+        
+        // left bound to constrain the right view to is either the far left (if no left view), or the left view
+        var leftBound = view
+        if let leftView = leftView {
+            view.addSubview(leftView)
+            leftView.leftAnchor.constraint(equalTo: view.leftAnchor).isActive = true
+            view.topAnchor.constraint(lessThanOrEqualTo: leftView.bottomAnchor).isActive = true
```
KevWho, line 156: It should be `leftView.topAnchor` instead of `leftView.bottomAnchor`, right?

bm-persona/Common/DetailView/OpenTimesCardView.swift
```diff
@@ -0,0 +1,196 @@
...
+                leftView.topAnchor.constraint(equalTo: view.topAnchor).isActive = true
+            }
+            leftBound = leftView
+        }
+        
+        if let rightView = rightView {
+            view.addSubview(rightView)
+            rightView.rightAnchor.constraint(equalTo: view.rightAnchor).isActive = true
+            rightView.leftAnchor.constraint(greaterThanOrEqualTo: leftBound.rightAnchor, constant: kViewMargin).isActive = true
+            view.topAnchor.constraint(lessThanOrEqualTo: rightView.bottomAnchor).isActive = true
```
KevWho, line 170: It should be `rightView.topAnchor` instead of `rightView.bottomAnchor`, right?

bm-persona/Common/DetailView/OpenTimesCardView.swift
```diff
@@ -0,0 +1,196 @@
...
+        
+        if let isOpen = item.isOpen {
+            if isOpen {
+                openTag.text = "Open"
+                openTag.backgroundColor = Color.openTag
+            } else {
+                openTag.text = "Closed"
+                openTag.backgroundColor = Color.closedTag
+            }
+            openTag.widthAnchor.constraint(equalToConstant: 50).isActive = true
```
KevWho, line 44: Is there a need to explicitly specify the width of the tag?

bm-persona/Libraries/LibraryDetailViewController.swift
```diff
@@ -49,156 +40,101 @@ class LibraryDetailViewController: SearchDrawerViewController {
...
-        scrollView.layoutMargins = UIEdgeInsets(top: 0, left: 10, bottom: 0, right: 10)
-        scrollView.alwaysBounceVertical = true
-        view.addSubview(scrollView)
-        scrollView.translatesAutoresizingMaskIntoConstraints = false
-        scrollView.topAnchor.constraint(equalTo: barView.topAnchor, constant: kViewMargin).isActive = true
-        scrollView.leftAnchor.constraint(equalTo: view.layoutMarginsGuide.leftAnchor).isActive = true
-        scrollView.rightAnchor.constraint(equalTo: view.layoutMarginsGuide.rightAnchor).isActive = true
-        scrollView.bottomAnchor.constraint(equalTo: view.layoutMarginsGuide.bottomAnchor).isActive = true
+        view.addSubview(scrollContainer)
+        scrollContainer.topAnchor.constraint(equalTo: barView.topAnchor, constant: kViewMargin).isActive = true
```
KevWho, line 124: Why did you choose to compare both top anchors instead of the vertical distance?

bm-persona/Libraries/LibraryDetailViewController.swift
```diff
@@ -9,31 +9,22 @@
 import UIKit
 import CoreLocation
 
-fileprivate let kHeaderFont: UIFont = Font.bold(24)
-fileprivate let kCardPadding: UIEdgeInsets = UIEdgeInsets(top: 5, left: 10, bottom: 5, right: 10)
+fileprivate let kCardPadding: UIEdgeInsets = UIEdgeInsets(top: 16, left: 16, bottom: 16, right: 16)
 fileprivate let kViewMargin: CGFloat = 16
 
 class LibraryDetailViewController: SearchDrawerViewController {
```
KevWho, line 15: I like what you've done with the place :)

bm-persona/Libraries/LibraryDetailViewController.swift
```diff
@@ -49,156 +40,101 @@ class LibraryDetailViewController: SearchDrawerViewController {
...
+        bookButton.rightAnchor.constraint(equalTo: scrollContent.rightAnchor).isActive = true
+        bookButton.bottomAnchor.constraint(equalTo: scrollContent.bottomAnchor).isActive = true
         bookButton.heightAnchor.constraint(equalToConstant: 50).isActive = true
-        bookButton.setTitle("Book a Study Room >", for: .normal)
-        bookButton.titleLabel!.font = Font.semibold(14)
-        bookButton.titleLabel!.textColor = .white
-        
     }
     
     func setUpScrollView() {
```
KevWho, line 122: I'm a bit confused about this hierarchy. So if I'm understanding this correctly, it goes `scrollContainer > scrollView > scrollContent`, with an additional `8px` padding between the `scrollView` and its content.

Could you explain a bit about the purpose of `scrollContainer` and why you chose a non-zero `scrollView.contentInset`?

bm-persona/Common/DetailView/OpenTimesCardView.swift
huangshawn1: ran into some situations where the tag was stretched and I guess for consistency. I can try getting rid of these width constraints and see what happens

bm-persona/Libraries/LibraryDetailViewController.swift
huangshawn1: was a mistake

bm-persona/Common/CollapsibleCardView.swift
```diff
@@ -0,0 +1,146 @@
...
+        UIView.animate(withDuration: 0.2) {
+            self.setState(opened: !self.isOpen)
+            if self.isOpen {
+                self.chevronIcon.transform = self.chevronIcon.transform.rotated(by: -1 * .pi / 2)
+            } else {
+                self.chevronIcon.transform = self.chevronIcon.transform.rotated(by: .pi / 2)
+            }
+        } completion: { _ in
+            completion?()
+        }
```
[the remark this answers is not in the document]
huangshawn1: wasn't getting the error because I was using Xcode 12 beta and I think that comes with a new version of swift where this syntax is allowed. changed it to work in Xcode 11 as well

bm-persona/Libraries/LibraryDetailViewController.swift
huangshawn1: so I wanted a margin of 16 between the scroll view and the card so that the user can move the drawer by dragging the left or right margin (otherwise the scroll would register as scrolling the scroll view). but I was running into some problem (I think the scroll view wasn't scrolling or something? don't remember exactly) where the solution was that the scroll view needed to be constrained with 0 margins to its superview. so I used scrollContainer to get the margins and constrained the scroll view to it with 0 margins. the contentInset is nonzero because otherwise the shadows would get cut off.

might be a better way to do this but this is the best I could come up with after running into some issues since I haven't used scroll views much.

bm-persona/Common/DetailView/OpenTimesCardView.swift
huangshawn1: got rid of it to use TagView.open and TagView.closed. don't see any issues so far

sidthesquid:
Beautiful

PR #184, opened 2020-08-16
Location Manager + Detail Consolidation

KevWho:
Adds `LocationManager`, a simple wrapper for `CLLocationManager` that fetches updates to the user's location and notifies observers. Using a singleton instance, we can significantly simplify location requests throughout the app and clean up some code.
- Adds `LocationManager` and `Notification.Name.locationUpdated`, which is used to notify listeners about a new location update.
- Removes `CLLocationManager` instances in other classes
- Changes `HasLocation` to use the new manager
  - `getDistanceToUser(userLoc:)` > `distanceToUser`, an optional boolean
  - Removes the `invalidDistance` double from models
  - Adds `locationComparator` and `locationFilter`, since they are commonly used for models conforming to `HasLocation`

Also consolidated how distance information is presented into `LocationDetailView`, which is the same walking man icon + distance label currently used in table cells and the overview card.
- Added `DetailView` and `DetailViewDelegate`, protocols for displaying details regarding a single protocol
  - I feel like these are kinda unnecessary, but I think it might be helpful in the future if we also consolidate the views for things like occupancy, open/closed, etc.?

Other changes
- Moved `UIView.iconPairView` from `OverviewCardView` to `Common/IconPairView.swift`
- Added `emptyOrChildrenHidden` bool to `UIStackView`

Curious to hear what people think of these changes. It doesn't change anything visually / functionally in the app, but the goal is to reduce some of the repeated blocks.

### DIFF
--- a/bm-persona/Common/DetailView/LocationDetailView.swift
+++ b/bm-persona/Common/DetailView/LocationDetailView.swift
@@ -1,0 +1,77 @@
+//
+//  LocationDetailView.swift
+//  bm-persona
+//
+//  Created by Kevin Hu on 4/24/20.
+//  Copyright © 2020 RJ Pimentel. All rights reserved.
+//
+
+import Foundation
+import UIKit
+
+/// The standard view for displaying the distance between the user and an item conforming to `HasLocation`.
+class LocationDetailView: IconPairView, DetailView {
+
+    typealias Item = HasLocation
+
+    private var icon: UIImageView!
+    private var label: UILabel!
+
+    /// The item configuring this view.
+    private var item: HasLocation?
+
+    weak var delegate: DetailViewDelegate?
+
+    var missingData: Bool {
+        guard let item = item else { return true }
+        return item.distanceToUser == nil
+    }
+
+    func configure(for item: HasLocation) {
+        self.item = item
+        label.text = "Unknonw"
+        if let distance = item.distanceToUser {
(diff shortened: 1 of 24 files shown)


bm-persona/Data/LocationManager.swift
```diff
@@ -0,0 +1,82 @@
...
+
+        _locationManager = CLLocationManager()
+        _locationManager.delegate = self
+        _locationManager.desiredAccuracy = kCLLocationAccuracyBest
+        _locationManager.requestWhenInUseAuthorization()
+        userLocation = _locationManager.location
+    }
+
+    deinit {
+        _locationManager.stopUpdatingLocation()
```
Stereopoly, line 48: Sick 💯

bm-persona/Common/DetailView/LocationDetailView.swift
```diff
@@ -0,0 +1,77 @@
...
+    weak var delegate: DetailViewDelegate?
+
+    var missingData: Bool {
+        guard let item = item else { return true }
+        return item.distanceToUser == nil
+    }
+
+    func configure(for item: HasLocation) {
+        self.item = item
+        label.text = "Unknonw"
```
huangshawn1, line 32: typo

KevWho: Fixed!

PR #185, opened 2020-08-29
Map Marker Updates

KevWho:
Updates Map Markers for new dataset.
- Add Cafe, Waste, Lactation, and Campus Garden as Map Markers. Updates for name changes (e.g. "Water Fountains" > "Water", "Ford Go Bikes" > "Lyft Bikes")
- Adds average meal price field for cafes. This is displayed in the bottom row of the detail view:
Screen Shot 2020-08-29 at 2 59 41 PM

## Known Issues
- Waiting on icons from design for the new categories. Current colors and icons are placeholders.
- Some of the data in the backend has "Not Available" in the names or in certain fields, which are reflected in the app.

## Future Work
- Transition to use the "Map Marker" collection in firebase instead of enumerating the available collections in the app.


bm-persona/Map/MapDataSource/MapMarker.swift
```diff
@@ -35,37 +39,63 @@ enum MapMarkerType: String, CaseIterable {
...
         case .bikes:
             return UIColor(displayP3Red: 45/255, green: 53/255, blue: 255/255, alpha: 1.0)
+        case .lactation:
+            return UIColor(displayP3Red: 249/255, green: 134/255, blue: 49/255, alpha: 1.0)
+        case .waste:
+            return UIColor(displayP3Red: 101/255, green: 54/255, blue: 17/255, alpha: 1.0)
+        case .garden:
+            return UIColor(displayP3Red: 124/255, green: 190/255, blue: 49/255, alpha: 1.0)
+        case .cafe:
+            return UIColor(displayP3Red: 146/255, green: 83/255, blue: 163/255, alpha: 1.0)
```
huangshawn1, line 98: do we want to move these to Colors.swift as well? could have a MapMarker struct inside the Color struct

bm-persona/Map/MapMarkerDetailView.swift
```diff
@@ -221,6 +226,16 @@ enum MapMarkerDetail {
             label.textColor = Color.primaryText
             label.text = description
             return viewWithIcon(icon, view: label)
+        case .price:
+            guard let price = marker.mealPrice else { return nil }
+            let icon = UIImage(named: "Dining")
```
huangshawn1, line 231: the icon looks kinda "squashed"/pixelated/cut off on my screen. the dining icon when you search a dining hall is pretty high-res, so I think this should be a solvable issue?
the clock and pin icons on the detail view also look a bit pixelated

bm-persona/Map/MapDataSource/MapMarker.swift
```diff
@@ -35,37 +39,63 @@ enum MapMarkerType: String, CaseIterable {
...
+            icon = UIImage(named: "lactation-icon")
+            break
+        case .waste:
+            icon = UIImage(named: "waste-icon")
+            break
+        case .garden:
+            icon = UIImage(named: "garden-icon")
+            break
+        case .cafe:
+            icon = UIImage(named: "cafe-icon")
```
huangshawn1, line 70: some of these icons look kinda pixelated on my phone, is there a reason for that?

bm-persona/Utils/UIImage+Extensions.swift
```diff
@@ -40,4 +40,35 @@ extension UIImage {
...
+        imageView.layer.masksToBounds = true
+        imageView.layer.borderWidth = width
+        imageView.layer.borderColor = color.cgColor
+        UIGraphicsBeginImageContextWithOptions(imageView.bounds.size, false, scale)
+        guard let context = UIGraphicsGetCurrentContext() else { return nil }
+        imageView.layer.render(in: context)
+        let result = UIGraphicsGetImageFromCurrentImageContext()
+        UIGraphicsEndImageContext()
+        return result
+    }
```
huangshawn1, line 61: this might be the cause of the icon issues?

bm-persona/Map/MapDataSource/MapMarker.swift
huangshawn1: IMG_1722

bm-persona/Map/MapMarkerDetailView.swift
huangshawn1: IMG_1723

huangshawn1 (changes requested):
some of the icons look pixelated. also noticed that if you search for something (e.g. dining hall) while map markers are up, the dining hall doesn't drop a pin (this may or may not be related to this pr?).

bm-persona/Map/MapMarkerDetailView.swift
KevWho: Fixed!

bm-persona/Map/MapDataSource/MapMarker.swift
KevWho: Fixed! Looks like it was a bug in `UIImage.resized(...)`

KevWho:
> also noticed that if you search for something (e.g. dining hall) while map markers are up, the dining hall doesn't drop a pin (this may or may not be related to this pr?).

Looks like the pin is not being displayed when there are other annotation views very close to the location. Fixed this by setting `displayPriority` to `.required`, so all annotation views will be displayed.

huangshawn1 (approved):
looks good

PR #193, opened 2020-08-31
Labels getting truncated - overview card

huangshawn1:
Overview card elements were getting truncated, especially on smaller phones.
Addressed by making labels 2 lines, cutting margins from 16 to 10.

Screenshots: first 3 iPhone SE, second 3 iPhone X

Simulator Screen Shot - iPhone SE (1st generation) - 2020-08-30 at 18 47 37
Simulator Screen Shot - iPhone SE (1st generation) - 2020-08-30 at 18 47 52
Simulator Screen Shot - iPhone SE (1st generation) - 2020-08-30 at 18 51 30
IMG_1641
IMG_1639
IMG_1640

### DIFF
--- a/bm-persona/Common/DetailView/LocationDetailView.swift
+++ b/bm-persona/Common/DetailView/LocationDetailView.swift
@@ -49,7 +49,7 @@ class LocationDetailView: IconPairView, DetailView {
         label.textColor = Color.blackText
         label.adjustsFontSizeToFitWidth = true
         label.minimumScaleFactor = 0.75
-        label.numberOfLines = 1
+        label.numberOfLines = 2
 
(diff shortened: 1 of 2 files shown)


huangshawn1:
> Nice job. Would there be any issues with setting the number of lines for some of the labels to 0? Specifically, for the address label, it's definitely likely that a long address could extend to > 2 lines. Wondering if this could cause visual bugs or if we'd prefer to truncate in that case.

Tested 0 lines for address and it worked well (takes more lines until no space, then truncates). I think this will work if one of the labels is 0 lines and others are defined, so I'll change address to 0 lines. Other attributes shouldn't take more than 2/3 lines anyway. If they're all 0 lines it would be ambiguous which label takes priority.

PR #197, opened 2020-09-05
Change "Nearby" button to sort

huangshawn1:
"Nearby" now sorts by distance instead of filtering by < 10 mi.
FilterTableView can now handle sorts as well as filters.
Default select both "nearby" and "open".
Fixed bug with Resources page (didn't use filtered data so filter didn't work).

Added:
- TableFunction protocol - includes sorts and filters
- Sort struct
- HasName protocol so that alphabetical sort works on anything with a name (previously operated on SearchItems but that didn't work with menu items)

### DIFF
--- a/bm-persona/Data/ItemProtocols/SearchItem.swift
+++ b/bm-persona/Data/ItemProtocols/SearchItem.swift
@@ -8,7 +8,7 @@
 
 import UIKit
 
-protocol SearchItem {
+protocol SearchItem: HasName {
 
     var searchName: String { get }
     var location: (Double, Double) { get }
--- a/bm-persona/Fitness/FitnessViewController.swift
+++ b/bm-persona/Fitness/FitnessViewController.swift
@@ -257,8 +257,8 @@ extension FitnessViewController {
     
     func setupFilterTableView() {
         let functions: [TableFunction] = [
-            Sort<Gym>(label: "Nearby", sort: Library.locationComparator()),
-            Filter<Gym>(label: "Open", filter: {lib in lib.isOpen ?? false}),
+            Sort<Gym>(label: "Nearby", sort: Gym.locationComparator()),
+            Filter<Gym>(label: "Open", filter: {gym in gym.isOpen ?? false}),
         ]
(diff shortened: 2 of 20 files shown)


bm-persona/Fitness/FitnessViewController.swift
```diff
@@ -257,11 +256,11 @@ extension FitnessViewController {
     }
     
     func setupFilterTableView() {
-        let filters = [
-            Filter<Gym>(label: "Nearby", filter: Gym.locationFilter(by: Gym.nearbyDistance)),
-            Filter<Gym>(label: "Open", filter: {gym in gym.isOpen ?? false}),
+        let functions: [TableFunction] = [
+            Sort<Gym>(label: "Nearby", sort: Library.locationComparator()),
+            Filter<Gym>(label: "Open", filter: {lib in lib.isOpen ?? false}),
```
KevWho, line 261: They refer to the same thing, but I think it should be `Gym.locationComparator()`. Another inconsequential thing, but I think for this view controller, it makes more sense if `gym` is the variable name instead of `lib`, haha.

Same goes for `DiningViewController`, I guess. Just with Dining Hall themed stuff instead.

KevWho (changes requested):
Nice 👍

bm-persona/Data/ItemProtocols/SearchItem.swift
```diff
@@ -8,7 +8,7 @@
 
 import UIKit
 
-protocol SearchItem {
+protocol SearchItem: HasName {
 
     var searchName: String { get }
```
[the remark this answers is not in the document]
huangshawn1: searchName is used in searchLocations in the MapViewController and it *might* be useful to keep a separate search name and "display" name for the future. could consolidate them and have them both be "name" too

bm-persona/Fitness/FitnessViewController.swift
huangshawn1: oh yeah that was copy and paste error, forgot to change

KevWho (approved):
Sweet

PR #198, opened 2020-09-06
Gym and Resource Detail Views

KevWho:
- Adds `ScrollingStackView`, a view with a `UIStackView` that scrolls vertically once the stack view contents exceed the bounds of the view.
  - Changes `LibraryDetailViewController` to use this new view and simplify some code
- Adds detail views for Gyms and Campus Resources using `ScrollingStackView`.
  - For Campus Resources, since these are on a separate tab, these present modally when a resource cell is clicked (instead of being brought up in a drawer).
- Shows detail view for Gyms and Campus Resources when searched.

Other Changes
- Fixes filters on Resources tab not doing anything.
- Adds "Nearby" filter to Resources tab.

There's still a fair amount of repeated code between all these detail views (all of them create a `ScrollingStackView` and add an overview card, for example), wondering if y'all think I should unify them some more (e.g. with a general `DetailViewController`), or if we should keep them separate for potential divergences in the future.

### DIFF
--- a/bm-persona/Common/ScrollingStackView.swift
+++ b/bm-persona/Common/ScrollingStackView.swift
@@ -65,7 +65,6 @@ class ScrollingStackView: UIView {
         stackView.bottomAnchor.constraint(equalTo: scrollView.bottomAnchor).isActive = true
         stackView.leftAnchor.constraint(equalTo: scrollView.layoutMarginsGuide.leftAnchor).isActive = true
         stackView.rightAnchor.constraint(equalTo: scrollView.layoutMarginsGuide.rightAnchor).isActive = true
-        stackView.widthAnchor.constraint(equalTo: scrollView.layoutMarginsGuide.widthAnchor).isActive = true
     }
 
 }
--- a/bm-persona/Common/ScrollingStackView.swift
+++ b/bm-persona/Common/ScrollingStackView.swift
@@ -1,0 +1,71 @@
+//
+//  ScrollingStackView.swift
+//  bm-persona
+//
+//  Created by Kevin Hu on 9/5/20.
+//  Copyright © 2020 RJ Pimentel. All rights reserved.
+//
+
+import UIKit
+
+/// A view containing a vertical `UIStackView` in a `UIScrollView` with a dynamic content height to match the contents of the stack.
+class ScrollingStackView: UIView {
+
+    /// The `UIScrollView` containing `stackView.
+    open var scrollView: UIScrollView!
+
+    /// The `UIStackView` containing the contents of this view.
+    open var stackView: UIStackView!
+
+    /// The vertical offset applied to the contents of `scrollView` relative to this view's parent.
+    open var yOffset: CGFloat {
+        return frame.minY + scrollView.contentInset.top
+    }
+
+    // MARK: Initializers
+
+    override init(frame: CGRect) {
+        super.init(frame: frame)
+        translatesAutoresizingMaskIntoConstraints = false
+        
+        setupScrollView()
+        setupStackView()
+    }
+
+    convenience init() {
+        self.init(frame: .zero)
+    }
+
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    // MARK: Views
+
+    func setupScrollView() {
+        scrollView = UIScrollView()
+        scrollView.contentInset = UIEdgeInsets(top: 8, left: 0, bottom: 8, right: 0)
+        scrollView.layoutMargins = UIEdgeInsets(top: 0, left: 8, bottom: 0, right: 8)
+        addSubview(scrollView)
+        scrollView.translatesAutoresizingMaskIntoConstraints = false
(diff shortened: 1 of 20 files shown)


bm-persona/Common/ScrollingStackView.swift
```diff
@@ -0,0 +1,71 @@
...
+        stackView.axis = .vertical
+        stackView.distribution = .fill
+        stackView.alignment = .fill
+        scrollView.addSubview(stackView)
+        stackView.translatesAutoresizingMaskIntoConstraints = false
+        stackView.topAnchor.constraint(equalTo: scrollView.topAnchor).isActive = true
+        stackView.bottomAnchor.constraint(equalTo: scrollView.bottomAnchor).isActive = true
+        stackView.leftAnchor.constraint(equalTo: scrollView.layoutMarginsGuide.leftAnchor).isActive = true
+        stackView.rightAnchor.constraint(equalTo: scrollView.layoutMarginsGuide.rightAnchor).isActive = true
+        stackView.widthAnchor.constraint(equalTo: scrollView.layoutMarginsGuide.widthAnchor).isActive = true
```
huangshawn1, line 68: is there a need to constrain left/right and width?

bm-persona/Resources/ResourceDetailViewController.swift
```diff
@@ -0,0 +1,93 @@
...
+        overviewCard = OverviewCardView(item: resource, excludedElements: [.openTimes])
+        overviewCard.heightAnchor.constraint(equalToConstant: 200).isActive = true
+        scrollingStackView.stackView.addArrangedSubview(overviewCard)
+    }
+
+    func setUpOpenTimesCard() {
+        guard resource.weeklyHours != nil else { return }
+        openTimesCard = OpenTimesCardView(item: resource, animationView: scrollingStackView, toggleAction: { open in
+            if open, self.currState != .full {
+                self.delegate.moveDrawer(to: .full, duration: 0.6)
```
huangshawn1, line 79: crashes on this line when open times card is opened because the delegate isn't set when presented modally.

bm-persona/Fitness/FitnessViewController.swift
```diff
@@ -14,8 +14,27 @@ fileprivate let kCardPadding: UIEdgeInsets = UIEdgeInsets(top: 16, left: 16, bot
 fileprivate let kViewMargin: CGFloat = 16
 fileprivate let kTodayClassesHeight: CGFloat = 300
 
-class FitnessViewController: UIViewController {
-    
+class FitnessViewController: UIViewController, SearchDrawerViewDelegate {
```
huangshawn1, line 17: could be moved to a future time, but should we add some sort of n/a on the open times card? right now, the gyms have no open times data (all blank) and it looks kinda weird. could also just not show it at all. I think there's a nil check but no empty check

bm-persona/Fitness/FitnessViewController.swift
```diff
@@ -14,8 +14,27 @@ fileprivate let kCardPadding: UIEdgeInsets = UIEdgeInsets(top: 16, left: 16, bot
...
+    // SearchDrawerViewDelegate property
+    var mainContainer: MainContainerViewController?
+
+    func handlePanGesture(gesture: UIPanGestureRecognizer) {
+        let state = handlePan(gesture: gesture)
+        if state == .hidden {
+            // get rid of the top detail drawer if user sends it to bottom of screen
+            mainContainer?.dismissTop()
+        }
+    }
```
huangshawn1, line 34: could move this logic into SearchDrawerViewDelegate

bm-persona/Libraries/LibraryDetailViewController.swift
```diff
@@ -28,48 +27,28 @@ class LibraryDetailViewController: SearchDrawerViewController {
...
-        let view = UIView()
-        view.translatesAutoresizingMaskIntoConstraints = false
-        return view
-    }()
-    var scrollContent: UIView = {
-        let view = UIView()
-        view.translatesAutoresizingMaskIntoConstraints = false
-        return view
+
+    var scrollingStackView: ScrollingStackView = {
```
huangshawn1, line 45: damn scrollingStackView really simplified/cleaned up this vc very nice

huangshawn1 (changes requested):
Yeah a general DetailViewController would probably be good. covers libraries, gyms, and resources and most detail views probably have an overview card and open times card anyway. unifies the code setting up those two cards, scrolling stack view, cutoff position, etc which is a significant amount of code. can also override stuff for changes in the future if necessary.

bm-persona/Fitness/FitnessViewController.swift
KevWho: Good point - added an empty check when parsing.

bm-persona/Common/ScrollingStackView.swift
KevWho: Doesn't look like it, removed the width constraint.

huangshawn1 (approved):
looks good, slight problem with open times card for resources

PR #199, opened 2020-09-12
Dining Detail View Fixes

Stereopoly:
* Filter out whitespace and items length < 1 from backend data (we had "." items and lots of "  Pancakes "
* Fix missing view from appearing after filtering everything out and switching

### DIFF
--- a/bm-persona/Dining/DiningDetailViewController.swift
+++ b/bm-persona/Dining/DiningDetailViewController.swift
@@ -144,7 +144,7 @@ extension DiningDetailViewController: TabBarControlDelegate {
         self.menuView.setData(data: meals[mealNames[control.index]]!)
         self.menuView.tableView.reloadData()
         self.menuView.filter.deselectAllItems()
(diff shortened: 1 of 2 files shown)


bm-persona/Dining/DiningDetailViewController.swift
```diff
@@ -144,6 +144,7 @@ extension DiningDetailViewController: TabBarControlDelegate {
         self.menuView.setData(data: meals[mealNames[control.index]]!)
         self.menuView.tableView.reloadData()
         self.menuView.filter.deselectAllItems()
+        self.menuView.missingView.isHidden = true
```
KevWho, line 147: What if you called `self.menuView.update()` instead? The missing view is updated in that function, and it could be the case that the tab you switch to still has no items, even if the filters are all deselected.

KevWho (changes requested):

Stereopoly:
Good idea - done!

KevWho (approved):
Nice

PR #201, opened 2020-09-12
Remove "Recommended" Label

KevWho:
Removes "Recommended" label in top right of cells for Libraries, Dining Halls, and Gyms. Leaves space for the label (only hides it), and top-aligns the name text.
Screen Shot 2020-09-12 at 4 02 31 PM
Screen Shot 2020-09-12 at 4 02 29 PM

Also fixes issue where Resource cells were taller than expected.


Stereopoly (approved):
Looks good

PR #202, opened 2020-09-14
Small Phone Layout Issues + Only default select "nearby"

huangshawn1:
Some text was truncated/overlapping on smaller phones.
- distance label on cells: no longer using stack, now left/right constrained
- certain library/resource names in cell and detail view: reduce minimum scale factor
- dining menu "late night": all segmented controls' labels auto-shrink to 0.6 if necessary
- "Today's classes" and "see full schedule" overlapped in fitness: fixed by label change to "Today", added a constraint anyway for future

Only default select "nearby" instead of both "nearby" and "open". shows the user the difference between selected and unselected filters.

Screenshots (SE):
Simulator Screen Shot - iPhone SE (1st generation) - 2020-09-13 at 17 53 36
Simulator Screen Shot - iPhone SE (1st generation) - 2020-09-13 at 18 03 37
Simulator Screen Shot - iPhone SE (1st generation) - 2020-09-13 at 17 53 49
Simulator Screen Shot - iPhone SE (1st generation) - 2020-09-13 at 18 03 19

### DIFF
--- a/bm-persona/Drawer/DrawerTabBar/TabBarControl.swift
+++ b/bm-persona/Drawer/DrawerTabBar/TabBarControl.swift
@@ -87,12 +87,25 @@ class TabBarControl: UISegmentedControl {
         segments = segments.filter { $0 != indicator }
         layoutIfNeeded()
         widths = []
+        setLabelsAdjustFont(view: self)
         for i in 0..<min(segments.count, numberOfSegments) {
-            widths.append(segments[i].frame.size.width)
+            widths.append(min(segments[i].frame.size.width, maxWidth / CGFloat(segments.count)))
             setWidth(maxWidth / CGFloat(segments.count), forSegmentAt: i)
         }
         layoutIfNeeded()
         updateIndicator()
+    }
+    
+    func setLabelsAdjustFont(view: UIView)  {
+        let subviews = view.subviews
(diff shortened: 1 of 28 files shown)


bm-persona/Drawer/DrawerTabBar/TabBarControl.swift
```diff
@@ -87,14 +87,27 @@ class TabBarControl: UISegmentedControl {
         segments = segments.filter { $0 != indicator }
         layoutIfNeeded()
         widths = []
+        setLabelsAdjustFont(view: self)
         for i in 0..<min(segments.count, numberOfSegments) {
-            widths.append(segments[i].frame.size.width)
+            widths.append(min(segments[i].frame.size.width, maxWidth / CGFloat(segments.count)))
```
KevWho, line 92: Wondering if you'd need to do the `min(..., maxWidth / CGFloat(segments.count))` if you moved the `layoutIfNeeded()` call to after you call `setLabelsAdjustFont(...)`.

KevWho (changes requested):

KevWho (approved):
Looks like my suggestion for `TabBarControl` didn't work! I've reverted your changes for that file to how they were before the revision.

PR #205, opened 2020-09-19
"Blobs"

Stereopoly:
| iPhone 8  | iPhone 8 Max   |  iPhone 11 Pro Max |
|---|---|---|
| Simulator Screen Shot - iPhone 8 - 2020-09-19 at 15 16 48 |  Simulator Screen Shot - iPhone 8 Plus - 2020-09-19 at 15 19 26| Simulator Screen Shot - iPhone 11 Pro Max - 2020-09-19 at 15 20 52 |
| Simulator Screen Shot - iPhone 8 - 2020-09-19 at 15 16 51 | Simulator Screen Shot - iPhone 8 Plus - 2020-09-19 at 15 19 29 | Simulator Screen Shot - iPhone 11 Pro Max - 2020-09-19 at 15 20 54 |


berkeley-mobile/Calendar/CalendarViewController.swift
```diff
@@ -214,3 +245,16 @@ extension CalendarViewController {
...
+        if #available(iOS 11.0, *) {
+            if UIApplication.shared.windows.count == 0 { return false }
+            let top = UIApplication.shared.windows[0].safeAreaInsets.top
+            return top > 20
+        } else {
+            // Fallback on earlier versions
+            return false
+        }
+    }
+}
```
KevWho, line 260: Could you move this to Utils, so it is easier to find for other classes?

berkeley-mobile/Calendar/CalendarViewController.swift
```diff
@@ -214,3 +245,16 @@ extension CalendarViewController {
         upcomingMissingView = MissingDataView(parentView: collectionView, text: "No upcoming events")
     }
 }
+
+extension UIDevice {
+    var hasNotch: Bool {
+        if #available(iOS 11.0, *) {
+            if UIApplication.shared.windows.count == 0 { return false }
+            let top = UIApplication.shared.windows[0].safeAreaInsets.top
```
KevWho, line 253: I know this is unlikely to cause any issues, but could you use `.first` instead of `[0]` and safely unwrap it? Just in case, haha.

KevWho (changes requested):
Couple visual issues with the 1st Gen SE.
Overlapping text. The "What do you need?" text is so long, blobs just might not be possible. If you wanted to add them, I'd set a constraint to enforce padding between the blob and the text, wrapping the text to two lines
Screen Shot 2020-09-19 at 3 48 51 PM
Gap at the top of the screen:
Screen Shot 2020-09-19 at 3 48 48 PM

berkeley-mobile/Calendar/CalendarViewController.swift
KevWho: Just noticed you check if `.count == 0` in the line above, so ignore this comment.

KevWho:
Looks like the vertical spacing issue was indeed an issue with the height of the image view being too large. I pushed a commit that constrains the aspect ratio of the image view to match that of the displayed image, and that seems to have fixed the issue.

It also looks like `.topAnchor` already refers to the top of the screen, below the status bar, so I fixed up some things regarding that constraint as well.

KevWho:
How it looks on SE (1st gen) now:
Screen Shot 2020-09-26 at 2 23 27 PM

Stereopoly:
Tested on all other devices bar se1

KevWho (approved):

PR #207, opened 2020-09-24
Hotfix: APM segfault

KevWho:
### Description

See [the issue on notion](https://www.notion.so/KERN_INVALID_ADDRESS-segfault-3ca79b7958424e32b36929009c2ca5c6) for crash details and investigation.

Addressed the issue by:
1. Running `pod update` to update Firebase (to `6.33.0`) and nanopb (to `1.30906.0`)
2. Clearing Firestore + Analytics persistent data when users first update to version `10.0.1` or later.
  (Added `AppDelegate.checkForUpdate()` that uses `UserDefaults` identify when the user first loads a new version of the app and perform necessary actions)

### Reviewing

Unfortunately, we don't have a way to reproduce the issue, so testing this fix is difficult.

As I updated a few large pods, most of the changed files are from running the update. The only changes made in `berkeley-mobile` are in:
- `berkeley-mobile.xcodeproj/project.pbxproj`
- `AppDelegate.swift`
- `AppDelegate+Migration.swift`


huangshawn1 (approved):
didn't seem to break anything. can't test if this fixed the issue without previous app version

PR #211, opened 2020-09-28
Calendar Tab View Height

KevWho:
Reduces the height of the tableviews in the Calendar tab to simplify user flow and make the view more intuitive.
Screen Shot 2020-09-27 at 10 26 49 PM


Stereopoly (approved):
Lgtm!

PR #215, opened 2020-10-10
Description Card

KevWho:
Adds additional card when viewing the details for Libraries, Gyms, and Campus Resources that displays a textual description of the resource. This card is inserted at the bottom of the page (or, in the case for Library, above the "Book a Study Room" button).
- Adds `DescriptionCardView`, a simple `CardView` subclass that has a label for a card title and a label to display some descriptive text, which syncs up presentation of the field across the different item types
- Adds optional `description` field to the models: `Library`, `Gym`, `Resource`
- Removes `description` parameter from the `SearchItem` protocol

Screen Shot 2020-10-10 at 12 32 45 AM

### Feedback Request

I used the mock-ups for the description card from the events page, but this looks a bit strange in the context of the detail pages for the Libraries + whatnot. My main complaint is with the card title:
- The title is a very large font, bigger than the title for the "Occupancy" card, and in some cases, the name of the item in the Overview card
- Is a title even necessary? Especially for items with shorter descriptions, it seems like we might be better off just displaying the description with no title:
Screen Shot 2020-10-10 at 12 40 34 AM

Also curious to hear any thoughts about the placement of the card within the page. I feel it might be better to have it above the occupancy card. What do y'all think?


Stereopoly:
Regarding position - it seems like most apps have it lower? Especially if its a long block of content. Also how much inset is there for the content in the card? Seems like each section has different insets which looks a bit off.

Stereopoly:
I also think the smaller font that Occupancy uses looks better than this large one and I prefer the positioning as well

KevWho:
Updated the font size and padding:
Screen Shot 2020-10-10 at 2 49 30 PM

KevWho:
Moved the description below the "Book Study Room" button, as discussed
Screen Shot 2020-10-11 at 1 36 41 PM

huangshawn1 (approved):
looks good, I like the description at the bottom below the button. personally prefer having no title instead of "Description" at the top but either way works

Stereopoly:
Looks good!

Stereopoly:
I think the title is good to have

PR #217, opened 2020-10-11
Adds "Learn More" button for Fitness location sites

KevWho:
## Description

Some fitness locations (e.g. hiking paths like Albany Bulb) have associated websites with additional information. Adds a "Learn More" button to the bottom of Gym details that opens these websites in Safari. This button is based off of the "Learn More" button from the event detail mock-up in Figma.
- Moves the implementation of the Library study room button to new `ActionButton`
- Adds optional `website` field to `Gym`
- Adds an action button to the bottom of the stack in `GymDetailViewController`
Screen Shot 2020-10-10 at 5 53 10 PM

## Testing

There is a [known issue](https://www.notion.so/Sprint-Board-iOS-3a44a1f5f7044adab2ea4b7827e7ea5c?p=20204d42576640d7910065740b374129) where the fields in Firestore have trailing / leading whitespace (e.g. the `"link"` field for Gyms is actually `" link"`.

As a workaround until this is fixed, to test this PR, replace the `"link"` key in [this line](https://github.com/asuc-octo/berkeley-mobile-ios/blob/a6b201ce278935706f32e4ed10d58956ce36d298/berkeley-mobile/Fitness/GymDataSource/GymDataSource.swift#L44) with `" link"`.

### DIFF
--- a/berkeley-mobile/Common/ActionButton.swift
+++ b/berkeley-mobile/Common/ActionButton.swift
@@ -1,0 +1,47 @@
+//
+//  ActionButton.swift
+//  berkeley-mobile
+//
+//  Created by Kevin Hu on 10/10/20.
+//  Copyright © 2020 ASUC OCTO. All rights reserved.
+//
+
+import UIKit
+
+/// The default padding to use for this button.
+fileprivate let kButtonPadding: UIEdgeInsets = UIEdgeInsets(top: 13, left: 16, bottom: 13, right: 16)
+
+/// A stylized button with a background color, rounded corners, and a drop shadow.
+class ActionButton: UIButton {
+
+    override open var isHighlighted: Bool {
+        didSet {
+            backgroundColor = isHighlighted ? Color.ActionButton.highlighted : Color.ActionButton.background
+        }
+    }
+
+    init(title: String, font: UIFont = Font.semibold(12)) {
+        super.init(frame: .zero)
+        contentEdgeInsets = kButtonPadding
+        isHighlighted = false
+
+        // Set rounded corners and drop shadow
+        layer.cornerRadius = 12
+        layer.shadowRadius = 5
+        layer.shadowOpacity = 0.25
+        layer.shadowOffset = .zero
+        layer.shadowColor = UIColor.black.cgColor
+        layer.shadowPath = UIBezierPath(rect: layer.bounds.insetBy(dx: 4, dy: 4)).cgPath
+
+        // Set title text and font
+        guard let label = titleLabel else { return }
+        setTitle(title, for: .normal)
+        label.font = font
+        label.textColor = Color.ActionButton.color
+    }
+
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+}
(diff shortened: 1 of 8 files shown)


berkeley-mobile/Fitness/GymDetailViewController.swift
```diff
@@ -38,6 +40,15 @@ class GymDetailViewController: SearchDrawerViewController {
         middleCutoffPosition = (openTimesCard?.frame.maxY ?? overviewCard.frame.maxY) + scrollingStackView.yOffset + 8
     }
 
+    /// Opens `gym.website` in Safari. Called as a result of tapping on `moreButton`.
+    @objc private func moreButtonClicked(sender: UIButton) {
+        guard let url = gym.website else { return }
+        presentAlertLinkUrl(title: "Are you sure you want to open Safari?",
+                            message: "Berkeley Mobile wants to open this resource's website",
```
huangshawn1, line 47: nitpick: do we want to say "resource's" here if we're opening a gym website? could also be the parks and stuff that backend added, we should probably figure out terminology. top of the card says "Fitness Centers" which isn't quite right either.
maybe "Fitness Location" or "Fitness Resource"?

huangshawn1 (commented):
works well, small nitpick with the message shown when opening the link

berkeley-mobile/Fitness/GymDetailViewController.swift
KevWho: Renamed the top of the card to say "Fitness Locations", and updated the alert message!

Stereopoly:
Booking library button has a ">" - thoughts?

Simulator Screen Shot - iPhone 11 Pro - 2020-10-11 at 23 33 57

KevWho:
> Booking library button has a ">" - thoughts?

It seems a bit unnecessary, but that's only if users can recognize the button as a button without it. Right now, it's just using the ">" character, if we wanted to keep it around it'd probably be better to make it match the arrow for the Open Hours card above. Here's how the button looks without the arrow, btw:
Screen Shot 2020-10-11 at 11 38 31 PM

It seems pretty to clear to me that it's interactable (with the blue background + action text), but I'm not really representative of a new user.

Stereopoly:
Yeah - I think we can remove the >

KevWho:
> Yeah - I think we can remove the >

Removed!

Stereopoly:
Awesome :)

berkeley-mobile/Common/ActionButton.swift
```diff
@@ -0,0 +1,47 @@
...
+
+/// The default padding to use for this button.
+fileprivate let kButtonPadding: UIEdgeInsets = UIEdgeInsets(top: 13, left: 16, bottom: 13, right: 16)
+
+/// A stylized button with a background color, rounded corners, and a drop shadow.
+class ActionButton: UIButton {
+
+    override open var isHighlighted: Bool {
+        didSet {
+            backgroundColor = isHighlighted ? Color.ActionButton.highlighted : Color.ActionButton.background
```
Stereopoly, line 19: Yayy it finally looks like you press it

Stereopoly (approved):

PR #218, opened 2020-10-17
Trim fields and strings

Stereopoly:
* Trim certain fields that commonly have spaces/newlines
* Replace � with an empty string, since it varies between apostrophe and other characters


huangshawn1 (approved):
looks good, can implement further string cleanup in future

PR #219, opened 2020-10-17
Drawer UX adjustments

Stereopoly:
* Adjust animations to align with iOS popovers and timing
* Test using real device since simulator lags hard
* Or use "Slow Animations" in simulator


huangshawn1 (commented):
looks good, I made some changes in the code to make sure all default drawer animation times were in the same place to make future changes easier. if someone could review/merge this that'd be great

Stereopoly (commented):
Looks good to me

KevWho (approved):
Looks good to me!

PR #220, opened 2020-10-17
Search Address for Gyms

KevWho:
Searching for Gyms always results in "Berkeley, CA" being displayed for the location name. This was inaccurate for resources like "Treasure Island".
- Change `SearchItem` to use `address` for `locationName`.


Stereopoly (approved):
Looks good to me and works!

huangshawn1 (approved):

PR #221, opened 2020-10-24
Safely index all TableViews

Stereopoly:
* Adds extension allowing for `safe` option which ensures index is valid (used in all TableViews now)
* Remove random academic folder


berkeley-mobile/Utils/Collection+Extension.swift
```diff
@@ -0,0 +1,17 @@
...
+//
+//  Created by Oscar Bjorkman on 10/24/20.
+//  Copyright © 2020 ASUC OCTO. All rights reserved.
+//
+
+import Foundation
+
+/** Ensures safety when indexing into an array. */
+extension Collection {
+    subscript(safe index: Index) -> Iterator.Element? {
```
KevWho, line 13: Nice

KevWho (changes requested):
Nice. Could we use this for `collectionView(_: cellForItemAt: ...)` as well?

KevWho (approved):
👍 Sick

PR #222, opened 2020-10-24
Apercu Font Change

KevWho:
Change font used in the app to Apercu (was previously Raleway).
- Remove `semibold` and `thin` font options and replace usages with `medium` and `light`.
- Change description text to `light` instead of `regular`.
Screen Shot 2020-10-24 at 3 06 09 PM
Screen Shot 2020-10-24 at 3 05 57 PM

Since `Apercu-Regular` is a bit heavier than `Raleway-Regular`, the search bar and results is a bit heavier than before. I can change this to `light`, depending on what y'all think:
Screen Shot 2020-10-24 at 3 07 22 PM


Stereopoly:
Looks better imo; old fonts were quite thin and light

Stereopoly (approved):

PR #223, opened 2020-10-24
Fitness Scrollview Height Fix

Stereopoly:
Title, reproduce issue using Network Conditioner to throttle network and quickly switch to Fitness tab


KevWho (approved):
Works for me!

PR #224, opened 2020-10-24
Campus Events Detail Page

huangshawn1:
Campus Events Detail View
- CampusEventDetailViewController: vc setting up page with overview card, description card, buttons to learn more/register
- EventOverviewCardView: overview card with time, location, name, favorite button
  - favorite button currently hidden

Other changes:
- removed upcoming cards. academic upcoming added back later: https://www.notion.so/Event-Upcoming-Cards-ccf500b7f3f749a6b34902b0f177966e
- added dateString getter for Calendar Events (format "Date / Time" or "Today / Time")
- removed if debug because backend has future events now

IMG_1770
IMG_1769
IMG_1768
IMG_1767
IMG_1766


berkeley-mobile/Calendar/Campus/CampusEventDetailViewController.swift
```diff
@@ -0,0 +1,100 @@
...
+        view.addSubview(scrollingStackView)
+        scrollingStackView.topAnchor.constraint(equalTo: barView.bottomAnchor, constant: kViewMargin).isActive = true
+        scrollingStackView.leftAnchor.constraint(equalTo: view.layoutMarginsGuide.leftAnchor).isActive = true
+        scrollingStackView.rightAnchor.constraint(equalTo: view.layoutMarginsGuide.rightAnchor).isActive = true
+        scrollingStackView.bottomAnchor.constraint(equalTo: view.layoutMarginsGuide.bottomAnchor).isActive = true
+    }
+    
+    func setUpOverviewCard() {
+        overviewCard = EventOverviewCardView(event: event)
+        overviewCard.heightAnchor.constraint(equalToConstant: 174).isActive = true
```
KevWho, line 50: Removing this constraint causes the overview card to grow in height dynamically depending on the event name. What do you think about this behavior? Would the inconsistency in size for this card for different events would be jarring?

KevWho (changes requested):
Looks good and works great! You were mentioning before that we shouldn't start loading the events lazily when the tab is selected, but I don't see that change here - did you change your mind?

berkeley-mobile/Calendar/Campus/CampusEventDetailViewController.swift
huangshawn1: instead of removing it I set a height anchor <= 200 so that it doesn't expand infinitely/too much even if backend sends some messed up/too long data. looks good with the resize so there's less empty space. almost all if not all names should fit.

huangshawn1: IMG_1791
IMG_1790

huangshawn1:
added EventsDataSource to the list of data sources loaded immediately. also made a slight change to split the list into normal data sources and those that must be loaded before occupancy can happen. this was to prevent occupancy from being very delayed because events takes a few seconds to load.

KevWho (approved):
Nice work!

PR #227, opened 2020-10-25
Fix load lag from data manager

huangshawn1:
Load events data source last to prevent it from happening at the same time as other data sources. This prevents too much lag initially before being able to see any dining halls/libraries.


KevWho (approved):
Thanks for the attention to detail!

PR #231, opened 2020-11-07
Update blobs to svgs

Stereopoly:
Blobs apart from those on launch screen now use svgs


huangshawn1 (approved):
looks good

KevWho:
Looks like the opacity issue was indeed a Xcode issue - merging!

PR #232, opened 2020-11-07
Eashan study pact onboarding

eashan-mathur:
Adds the initial onboarding flow to StudyPact.


Stereopoly:
Looks really good!
A small bug:
* It seems like if you press "next" to go through all the pages the "Let's get started" button never appears

PageViewController.swift
```diff
@@ -0,0 +1,211 @@
...
+        closeButton.setTitleColor(.gray, for: .normal)
+        closeButton.frame = CGRect(x: 30, y: 100, width: 50, height: 30)
+        view.addSubview(closeButton)
+        closeButton.addTarget(self, action: #selector(self.close(_:)), for: .touchUpInside)
+        closeButton.translatesAutoresizingMaskIntoConstraints = false
+        closeButton.leftAnchor.constraint(equalTo: view.safeAreaLayoutGuide.leftAnchor, constant: 24).isActive = true
+        closeButton.topAnchor.constraint(equalTo: view.safeAreaLayoutGuide.topAnchor).isActive = true
+        
+        
+        getStarted.setImage(UIImage(named: "getStarted"), for: .normal)
```
Stereopoly, line 80: I think we should just make this button in code instead of using the image, since it just has text and rounded corners

PageViewController.swift
```diff
@@ -0,0 +1,211 @@
...
+        self.pages.append(page3)
+        setViewControllers([pages[initialPage]], direction: .forward, animated: true, completion: nil)
+        
+        // pageControl
+        self.pageControl.frame = CGRect(x: 0, y: 0, width: 100, height: 100)
+        self.pageControl.currentPageIndicatorTintColor = UIColor.black
+        self.pageControl.pageIndicatorTintColor = UIColor.lightGray
+        self.pageControl.numberOfPages = self.pages.count
+        self.pageControl.currentPage = initialPage
+        self.pageControl.pageIndicatorTintColor = #colorLiteral(red: 0.337254902, green: 0.4352941176, blue: 0.7254901961, alpha: 1)
```
Stereopoly, line 138: Would be nice to create a swift file in the Assets/Colors and store these there

StudyPactOnboardingViewController.swift
```diff
@@ -0,0 +1,167 @@
...
+        signUpLabel.centerYAnchor.constraint(equalTo: self.view.centerYAnchor, constant: 65).isActive = true
+        
+        numberImage.image = UIImage(named: "Onboarding1Num")
+        numberImage.frame = CGRect(x: 0, y: 0, width: 34, height: 29)
+        numberImage.translatesAutoresizingMaskIntoConstraints = false
+        numberImage.rightAnchor.constraint(equalTo: signUpLabel.leftAnchor, constant: -20).isActive = true
+        numberImage.centerYAnchor.constraint(equalTo: signUpLabel.centerYAnchor).isActive = true
+        
+        descriptionText.text = "Complete your profile in less than 5 minutes to get the best pairing"
+        descriptionText.font = UIFont(name: "Raleway-Regular", size: 18)
```
Stereopoly, line 50: Use Fonts.regular(18)

StudyPactOnboardingViewController.swift
```diff
@@ -0,0 +1,167 @@
...
+        let numberImage = UIImageView()
+        let descriptionText = UILabel()
+        self.view.addSubview(signUpLabel)
+        self.view.addSubview(bearImage)
+        self.view.addSubview(numberImage)
+        self.view.addSubview(descriptionText)
+        
+        
+        bearImage.image = UIImage(named: "Onboarding2")
+        bearImage.frame = CGRect(x: 0, y: 0, width: 309, height: 321)
```
Stereopoly, line 83: Just commenting on one image - Probably better to constrain the image + label + description in a stackview or smt since the items don't dynamically adjust to screen size right now

PageViewController.swift
```diff
@@ -0,0 +1,211 @@
...
+        if currentIndex != pages.count - 1 {
+            setViewControllers([pages[currentIndex + 1]], direction: .forward, animated: true, completion: nil)
+        }
+        pageControl.currentPage += 1
+        
+    }
+    
+    func configurePageControl() {
+        let initialPage = 0
+        let page1 = StudyPactOnboardingViewController()
```
Stereopoly, line 122: Since each of the onboardings view are very similar, maybe just create a singular one of these with an init that has parameters like BearImage, Subtitle, etc.

Stereopoly (changes requested):

PageViewController.swift
```diff
@@ -0,0 +1,211 @@
...
+    }
+    
+    override func viewDidLayoutSubviews() {
+        pageControl.subviews.forEach {
+            $0.transform = CGAffineTransform(scaleX: 1.5, y: 1.5)
+        }
+    }
+    
+    func configureButtons() {
+        print("Button should show")
```
huangshawn1, line 46: remove print before merge

PageViewController.swift
```diff
@@ -0,0 +1,211 @@
...
+        let nextButton = UIButton()
+        nextButton.setTitle("Next", for: .normal)
+        nextButton.setTitleColor(.gray, for: .normal)
+        nextButton.frame = CGRect(x: 30, y: 100, width: 50, height: 30)
+        view.addSubview(nextButton)
+        nextButton.addTarget(self, action: #selector(nextClicked(_:)), for: .touchUpInside)
+        nextButton.translatesAutoresizingMaskIntoConstraints = false
```
huangshawn1, line 53: just a consistency thing but it might be better to create the button like you did with pageControl. same for the other buttons (and generally any other ui elements that are just initialized once). something like:
let nextButton: UIButton = {
        let button = UIButton()
        button.setTitle("Next", for: .normal)
        button.setTitleColor(.gray, for: .normal)
        button.frame = CGRect(x: 30, y: 100, width: 50, height: 30)
        button.addTarget(self, action: #selector(nextClicked(_:)), for: .touchUpInside)
        button.translatesAutoresizingMaskIntoConstraints = false
return button
}()

PageViewController.swift
```diff
@@ -0,0 +1,211 @@
...
+        nextButton.centerYAnchor.constraint(equalTo: pageControl.centerYAnchor).isActive = true
+        nextButton.rightAnchor.constraint(equalTo: view.safeAreaLayoutGuide.rightAnchor, constant: -24).isActive = true
+        nextButton.bottomAnchor.constraint(equalTo: view.safeAreaLayoutGuide.bottomAnchor).isActive = true
+        
+        
+        skipButton.setTitle("Skip", for: .normal)
+        skipButton.setTitleColor(.gray, for: .normal)
+        skipButton.frame = CGRect(x: 30, y: 100, width: 50, height: 30)
```
huangshawn1, line 61: generally don't know if it's a good idea to mix setting the frame and setting constraints. it looks like the x/y for the button are overridden anyway so the frame is just used to set the width/height. in that case, just setting width/height using constraints should achieve the same effect.
same for the other buttons

PageViewController.swift
```diff
@@ -0,0 +1,211 @@
...
+        
+        getStarted.setImage(UIImage(named: "getStarted"), for: .normal)
+        getStarted.frame = CGRect(x: 30, y: 100, width: 186, height: 43)
+        view.addSubview(getStarted)
+        getStarted.addTarget(self, action: #selector(self.close(_:)), for: .touchUpInside)
+        getStarted.translatesAutoresizingMaskIntoConstraints = false
+        getStarted.centerXAnchor.constraint(equalTo: view.safeAreaLayoutGuide.centerXAnchor).isActive = true
+        getStarted.bottomAnchor.constraint(equalTo: view.safeAreaLayoutGuide.bottomAnchor, constant: -130).isActive = true
+        getStarted.isHidden = true
+        getStarted.layer.opacity = 0
```
huangshawn1, line 88: is there a reason to do both? seems like just setting opacity should be enough?

PageViewController.swift
```diff
@@ -0,0 +1,211 @@
...
+        
+        for i in currentIndex...pages.count-1 {
+            setViewControllers([pages[i]], direction: .forward, animated: true, completion: nil)
+        }
+        pageControl.currentPage = pages.count - 1
+        
+    }
+    
+    @objc func nextClicked(_: UIButton){
+        print("Nexting now")
```
huangshawn1, line 107: remove

PageViewController.swift
huangshawn1: see for example Colors+Text.swift, Colors+Event.swift etc

PageViewController.swift
```diff
@@ -0,0 +1,211 @@
...
+        // add the individual viewControllers to the pageViewController
+        self.pages.append(page1)
+        self.pages.append(page2)
+        self.pages.append(page3)
+        setViewControllers([pages[initialPage]], direction: .forward, animated: true, completion: nil)
+        
+        // pageControl
+        self.pageControl.frame = CGRect(x: 0, y: 0, width: 100, height: 100)
+        self.pageControl.currentPageIndicatorTintColor = UIColor.black
+        self.pageControl.pageIndicatorTintColor = UIColor.lightGray
```
huangshawn1, line 135: figma seems to have white for non-current page circle and blue for current page circle. not a big deal though if there's a problem with doing it that way

PageViewController.swift
```diff
@@ -0,0 +1,211 @@
...
+        self.pageControl.bottomAnchor.constraint(equalTo: self.view.bottomAnchor, constant: -65).isActive = true
+        self.pageControl.widthAnchor.constraint(equalToConstant: 140).isActive = true
+        self.pageControl.heightAnchor.constraint(equalToConstant: 20).isActive = true
+        self.pageControl.centerXAnchor.constraint(equalTo: self.view.centerXAnchor).isActive = true
+        self.pageControl.tintColor = UIColor(displayP3Red: 86/255, green: 111/255, blue: 185/255, alpha: 1)
+        self.pageControl.isUserInteractionEnabled = false
+    }
+    
+    func pageViewController(_ pageViewController: UIPageViewController, viewControllerBefore viewController: UIViewController) -> UIViewController? {
+        if let viewControllerIndex = self.pages.firstIndex(of: viewController) {
```
huangshawn1, line 151: `guard let viewControllerIndex = self.pages.firstIndex(of: viewController), viewControllerIndex != 0 else { return nil }` instead?

PageViewController.swift
```diff
@@ -0,0 +1,211 @@
...
+        }
+        return nil
+    }
+    
+
+    
+    func pageViewController(_ pageViewController: UIPageViewController, viewControllerAfter viewController: UIViewController) -> UIViewController? {
+        if let viewControllerIndex = self.pages.firstIndex(of: viewController) {
+            
+            if viewControllerIndex < self.pages.count - 1 {
```
huangshawn1, line 170: guard let viewControllerIndex = self.pages.firstIndex(of: viewController), viewControllerIndex < self.pages.count - 1 else { return nil }

just cause its a bit cleaner

PageViewController.swift
```diff
@@ -0,0 +1,211 @@
...
+            return nil
+    }
+    
+    func pageViewController(_ pageViewController: UIPageViewController, didFinishAnimating finished: Bool, previousViewControllers: [UIViewController], transitionCompleted completed: Bool) {
+            
+    // set the pageControl.currentPage to the index of the current viewController in pages
+        if let viewControllers = pageViewController.viewControllers {
+            if let viewControllerIndex = self.pages.firstIndex(of: viewControllers[0]) {
+                self.pageControl.currentPage = viewControllerIndex
+                print(viewControllerIndex)
```
huangshawn1, line 185: remove

PageViewController.swift
```diff
@@ -0,0 +1,211 @@
...
+                if viewControllerIndex == self.pages.count - 1 {
+                    getStarted.isHidden = false
+                    UIView.animate(withDuration: 0.5) {
+                        self.getStarted.layer.opacity = 1
+                        self.skipButton.layer.opacity = 0
+                    }
+                    
+                }
+                
+                else if viewControllerIndex != self.pages.count - 1 {
```
huangshawn1, line 196: style thing but we generally have the else on the same line as the bracket

StudyPactOnboardingViewController.swift
```diff
@@ -0,0 +1,167 @@
...
+
+    let pageNumber = 0
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        view.backgroundColor = .white
+        let signUpLabel = UILabel()
+        let bearImage = UIImageView()
+        let numberImage = UIImageView()
+        let descriptionText = UILabel()
```
huangshawn1, line 22: try to initialize these using a
`let signUpLabel: UILabel = { configure and return the label }()`
style for consistency.
check out OverviewCardView lines 211-366 for reference

StudyPactOnboardingViewController.swift
huangshawn1: adding onto this, for all colors and fonts try to use Fonts or Colors. if the font/color you want isn't already somewhere in the app, you can add the new font/color yourself

huangshawn1 (changes requested):
just a few style/consistency things. also agree with oscar that the onboarding pages would be better as a single class with inputs like image, text, index, etc.
generally:

- try to create/configure ui elements using a `let button: UIButton = { }()` format for consistency and to kind of separate out the code to be more readable. check out OverviewCardView lines 211-366 for reference
- try not to set the frame. we generally use constraints instead to dynamically resize things
- remove prints
- we have a struct Font and struct Color for all the colors/fonts in the app
- using guard let ... { return some default value } is usually preferable

StudyPactOnboardingViewController.swift
eashan-mathur: Do you mean place the items inside of a stack view, then constrain the stack view in the view, and then constrain the individual things inside the stack view?

PageViewController.swift
eashan-mathur: This is needed to advance the dots

huangshawn1:
made some slight changes - got rid of some frame stuff in favor of constraints, the two added files were in the root directory so I moved them into berkeley-mobile/StudyPact directory, added `#if DEBUG` for bear button

huangshawn1 (approved):
looks good, I like the changes to generalize the study pact onboarding pages

Stereopoly (commented):
New changes look great, just a few screenshots below regarding the overall layout

Stereopoly:
Simulator Screen Shot - iPhone 12 Pro - 2020-12-05 at 16 05 55
Hide next button on final screen

Stereopoly:
image
Slightly too low labels

huangshawn1:
Simulator Screen Shot - iPhone SE (1st generation) - 2021-01-23 at 17 15 34
Simulator Screen Shot - iPhone SE (1st generation) - 2021-01-23 at 16 44 45

Stereopoly (changes requested):
Looks good all around - just one small position adjustment. Could you position the X so that it is equidistant from the left margin and the top margin? It looks funky on the phones without the notch.
image

Stereopoly (approved):
Fixes look good!

PR #234, opened 2020-11-14
Vertical Scrolling in a Drawer

KevWho:
If a drawer was opened, and the contents could be scrolled through vertically, it was difficult to dismiss the drawer, since gestures within the contents of the scrollview would move the scrollview, and not the drawer.

This PR changes this so that if the scrollview is at the top of its contents, dragging down will move the drawer, instead of the contents of the scrollview. Similarly, if the scrollview is at the bottom of its contents, or if the drawer is not fully extended, dragging up will move the drawer up.
- Adds `SimultaneousGestureScrollView`,  a `UIScrollView` subclass and replace the scrollview in `ScrollingStackView` with this new class.
- Adds a couple lines in each of `GymDetailViewController`, `LibraryDetailViewController`, and `ResourceViewController` to setup the scrollview
- Disables scrollview bouncing for `ScrollingStackView`

Let me know if any of these changes feel off!


huangshawn1 (commented):
scrolling to the top of the scroll view, having it suddenly stop, then having to pick up your finger and scroll again to dismiss feels a little off. I think it would feel better if it went smoothly from scrolling to the edge of the scroll view and pulling the drawer down (might be harder to implement though?).
also the thing oscar mentioned about scrolling down quickly feels off

Stereopoly:
It works pretty well except I noticed that when you open up a page like Moffett library that is longer than the screen, when you first press to scroll there is a snapping motion and the entire drawer moves up to the top. Also would it be possible to remove the ability to move the drawer past the top pos or add an animation for it, since theres some weird interactions with that too.

Stereopoly:
Also I think removing the "bounce" from the scroll view makes the other non-drawer pages feel pretty weird. Eg. the covid resource page

KevWho:
@Stereopoly I pushed a fix for the "snapping" behavior you were talking about. Also re-enabled bouncing, let me know if you prefer the feel of bouncing or no bouncing. With bouncing, the drawer is movable once the scrollview is done bouncing.

Stereopoly:
Love the changes! Feels really nice to me

huangshawn1 (changes requested):
bug where if you swipe up when the scroll view is at the top, instead of scrolling down the entire drawer is pulled up

huangshawn1 (approved):
looks good

PR #235, opened 2020-11-14
Set initial Dining Menu based on time

Stereopoly:
* Uses current time to select Breakfast, Lunch, Dinner, etc.
* Defaults to first item if time is not during a current meal


huangshawn1 (changes requested):
should sort the weeklyHours when first parsing them to ensure this works (don't think they're sorted rn/at first glance but could be wrong)

huangshawn1:
closing because if I remember correctly the current backend setup won't allow this sort of change

PR #236, opened 2020-11-27
Add booking section

Stereopoly:
* Adds booking section to Covid page
* Moves Covid to default and left-most resource page in segmented control
* Update displayed statistics

Simulator Screen Shot - iPhone 8 - 2020-11-26 at 17 53 05

### DIFF
--- a/berkeley-mobile/Resources/Covid/CovidResourceViewController.swift
+++ b/berkeley-mobile/Resources/Covid/CovidResourceViewController.swift
@@ -22,6 +22,7 @@ class CovidResourceViewController: UIViewController {
     private var lastUpdated: UILabel!
     
     private var screeningUrl = "https://calcentral.berkeley.edu/"   // fallback url
+    private var bookingUrl = "https://etang.berkeley.edu/appointments_home.aspx"
     
     override func viewDidLoad() {
         super.viewDidLoad()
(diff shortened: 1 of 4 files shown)


berkeley-mobile/Resources/Covid/CovidResourceViewController.swift
```diff
@@ -183,6 +183,65 @@ extension CovidResourceViewController {
...
+        card.addSubview(bookButton)
+        
+        bookButton.translatesAutoresizingMaskIntoConstraints = false
+        bookButton.topAnchor.constraint(equalTo: subtitleLabel.bottomAnchor, constant: kViewMargin).isActive = true
+        bookButton.leftAnchor.constraint(equalTo: card.layoutMarginsGuide.leftAnchor).isActive = true
+        
+        bookButton.setHeightConstraint(33)
+        bookButton.setWidthConstraint(110)
+        
+        card.bottomAnchor.constraint(equalTo: bookButton.bottomAnchor, constant: kViewMargin).isActive = true
```
KevWho, line 242: Would have constraining the `bookButton.bottomAnchor.constraint(equalTo: card.layoutMarginsGuide.bottomAnchor).isActive = true` done the same thing? If so, I think we should use that instead since it uses `kCardPadding` at the top of this function.

KevWho (changes requested):

KevWho (approved):
Awesome!

PR #237, opened 2020-12-05
Covid-19 Conflicting Constraints

KevWho:
Fixes conflicting constraints on the Covid-19 page. Removes the left/right constraints on the cards that have been added to the scrolling stack, and instead adjusts the margins / padding of the scrolling stack.


huangshawn1 (approved):
looks good, not seeing conflicts anymore

PR #240, opened 2020-12-06
Campus Resource Tags

KevWho:
Add the `tag` field from Firebase to the campus-wide resources, and display tags with an associated color:
Screen Shot 2020-12-05 at 5 17 50 PM
- Adds `ResourceType.swift` for known values ("Health", "Finances", "Legal", "Basic Needs", "Admin") and colors. Unknown values will be displayed with the default, gray color.


Stereopoly (approved):
Looks good!

PR #242, opened 2021-01-23
Study Pact Group Pages UI

huangshawn1:
Study Pact:
- change "Libraries" tab to "Study", renamed class
  - add Study Groups card above libraries card
  - cards are in ScrollView
- StudyGroup, StudyGroupMember structs
- StudyGroupsView: CollectionView arranging all study groups in a grid, with optional "Create Group" cell
  - StudyGroupCell class - cell with class name, profile pictures, attribute tags
    - gray and "Pending" shown if group.pending is true
  - CreateGroupCell class - dotted line border, "Create Group" button
- StudyPreferencesViewController: allows users to modify existing groups, delete groups

Other changes:
- CardTableViewCell class to generalize rounded cards with shadows used for table view cells
- function to select the profile tab in the tab bar

Modifications I made compared to Figma:
- removed the term "Study Pact" from text shown to users: from my understanding this is an internal project name?
- instead of "Get Started with Study Pact!" I used "Sign In to Get Started!" to apply to new users and old users who have been logged out
- instead of "Update Study Pact Preferences" button when the user hasn't joined any groups, the "Create Group" cell is shown if the user is in < 2 groups
- instead of red circles I used the profile picture icon for users with no photo
- got rid of "Add New Class Preference" on preferences page. users should instead use the "Create Group" cell
- instead of "Zoom" I used "Virtual" and "In Person" to match the creation flow

Notes:
- using dummy data to test UI
- whether the user is signed in (determines whether they are prompted to sign in or if groups are shown) is determined by the time for testing purposes
- works on small phones (SE 1st gen) from my testing, may want to verify

Screenshots:
IMG_1917
IMG_1914
IMG_1913
IMG_1916
IMG_1912
IMG_1915


berkeley-mobile/StudyPact/StudyGroups/AllStudyGroupsViewController.swift
```diff
@@ -0,0 +1,91 @@
...
+        vc.studyGroups = studyGroups
+        self.present(vc, animated: true, completion: nil)
+    }
+    
+    // MARK: Setup
+    override func viewDidLoad() {
+        super.viewDidLoad()
+
+        // Do any additional setup after loading the view.
+        setUpBackgroundView()
```
Stereopoly, line 51: I think most places use setup

Stereopoly (approved):
looks good!

PR #243, opened 2021-01-23
Add Profile Page

Stereopoly:
image


huangshawn1 (commented):
instead of OB profile picture, we should use the icon from the Profile tab in the tab bar as the default placeholder. we also want an "Edit Profile Picture" button (as on Figma) which opens the user's photos to let them choose a photo.
I think the vertical spacing between the profile picture and the first textfield should be increased a bit.
Link FB profile button will probably end up being a textfield instead.
also 21,000 lines lmao

berkeley-mobile/StudyPact/ProfileViewController.swift
```diff
@@ -0,0 +1,222 @@
...
+        let aspectRatio = blob.size.width / blob.size.height
+        let blobView = UIImageView(image: blob)
+        blobView.contentMode = .scaleAspectFit
+
+        view.addSubview(blobView)
+        blobView.translatesAutoresizingMaskIntoConstraints = false
+        blobView.widthAnchor.constraint(equalTo: view.widthAnchor, multiplier: 1 / 1.3).isActive = true
+        blobView.heightAnchor.constraint(equalTo: blobView.widthAnchor, multiplier: 1 / aspectRatio).isActive = true
+        blobView.topAnchor.constraint(equalTo: view.topAnchor, constant: -blobView.frame.height / 3.5).isActive = true
+        blobView.rightAnchor.constraint(equalTo: view.rightAnchor, constant: 40).isActive = true
```
huangshawn1, line 64: this blob code should probably be unified somewhere (maybe a class thats like BlobBackgroundView) because it's used multiple places and we will likely use it on future pages.

huangshawn1 (changes requested):
one minor code refactor change and what I mentioned in the above comment

berkeley-mobile/StudyPact/ProfileViewController.swift
Stereopoly: I'll leave this for another PR

Stereopoly:
* Added ability for user to set custom profile image (awaiting backend connection however)
* Initials used as fallback when no image is set
image

Stereopoly:
* Updated profile page
* Moved circular label out to own class
* Hotfix for font on onboarding page

huangshawn1 (approved):

PR #245, opened 2021-01-26
Split Resources Tab

KevWho:
Splits the Campus-Wide tab of the Resources tab into three distinct tags: `Basic Needs`, `Health`, `Admin` (in that order). Would like feedback about this change:
- Ordering of the categories.
- Would we want to not hardcode these categories in the case that new ones are added to the backend, or is that not a concern and having deliberate ordering is preferable?
- How does the scrolling of the segmented control feel / is it intuitive, or will many users not know that there may be obscured options? On the iPhone 12 Pro, `Admin` is completely cut off, and I'm not sure how to indicate to users that there are more pages.

## Description

Since the additional buttons flow off the edge of the screen for small screens, users can scroll through the options of the `SegmentedControl` horizontally.
- Adds the public `sizeEqually` and `spacing` variables to `SegmentedControl`.
  - When `sizeEqually` is `false`, the control will size the widths of the segments based on their text lengths, rather than having equal-sized segments that shrink the font size of longer labels when there is no room (previous behavior).
  - The `spacing` value is only used when `sizeEqually` is `false`, and determines the margins between segments.
- If the user swipes to a page whose label is cut-off in the segmented control, automatically scroll the segmented control to display the full text.

Splits the Resources tab into `Basic Needs`, `Health`, and `Admin`.
- Removes the `.finances` and `.legal` types from the `ResourceType` enum.
- Changes `CampusResourceViewController` to display only the filtered resources of a single type.

## After

Screen Shot 2021-01-25 at 10 07 15 PM

### DIFF
--- a/berkeley-mobile/Resources/ResourcesDataSource/ResourceType.swift
+++ b/berkeley-mobile/Resources/ResourcesDataSource/ResourceType.swift
@@ -12,8 +12,6 @@ import UIKit
 enum ResourceType: String {
 
     case health = "Health"
-    case finances = "Finances"
-    case legal = "Legal"
     case basicNeeds = "Basic Needs"
     case admin = "Admin"
 
@@ -21,10 +19,6 @@ enum ResourceType: String {
         switch self {
         case .health:
             return Color.Resource.health
-        case .finances:
-            return Color.Resource.finances
-        case .legal:
-            return Color.Resource.legal
         case .basicNeeds:
             return Color.Resource.basicNeeds
         case .admin:
--- a/berkeley-mobile/Assets/Colors/Colors.swift
+++ b/berkeley-mobile/Assets/Colors/Colors.swift
@@ -79,5 +79,9 @@ struct Color {
     static var barGraphEntryCurrent: UIColor {
         return UIColor(displayP3Red: 119/255, green: 154/255, blue: 252/255, alpha: 1.0)
     }
+
+    static var segmentedControlHighlight: UIColor {
+        return UIColor(displayP3Red: 250/255, green: 212/255, blue: 126/255, alpha: 1.0)
+    }
 }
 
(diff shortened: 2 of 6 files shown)


berkeley-mobile/Resources/ResourcesDataSource/ResourceType.swift
```diff
@@ -12,19 +12,13 @@ import UIKit
 enum ResourceType: String {
 
     case health = "Health"
-    case finances = "Finances"
-    case legal = "Legal"
     case basicNeeds = "Basic Needs"
     case admin = "Admin"
```
huangshawn1, line 16: I'd consider adding an unknown/other category in case of future changes. that way, everything backend has (including any new resources) will show up even if not categorized correctly.

huangshawn1 (commented):
personally, I would put basic needs last and maybe do (Health, Admin, Basic Needs) based on what I think the majority of students would use the most.
the cutoff for the scrolling categories seems a bit too far to the left, it looks a bit awkward to me. can't think of any better way to indicate scrollable other than having one of the labels be cut off.
IMG_1922

KevWho:
> personally, I would put basic needs last and maybe do (Health, Admin, Basic Needs) based on what I think the majority of students would use the most.

I based the ordering off of this, but I can change it around.

Screen Shot 2021-01-25 at 10 42 51 PM

Stereopoly:
I think some sort of fading action on the left and right edges of the segmented control could help show that it's movable

KevWho:
Pushed some changes: added a right inset to the controls, reorganized the tabs to `Health`, `Admin`, `Basic Needs`, and added an `Other` tab.

Also added a button to the right of the controls that is displayed when the user is not on the last tab and when the user is not scrolled all the way to the right:
Screen Shot 2021-01-30 at 3 32 22 PM

huangshawn1 (approved):
looks good, shadow looks a bit strange on the arrow button

KevWho:
w/o shadow:
Screen Shot 2021-01-30 at 3 46 55 PM

KevWho:
w/ same shadow as `CardView`:
Screen Shot 2021-01-30 at 3 54 15 PM

PR #246, opened 2021-01-30
StudyPact group view design changes

huangshawn1:
Changed some elements of the study group views to match the updated design on Figma: https://www.figma.com/file/24WqYuj4DH8Fjn3yRwuJtk/Berkeley-Mobile?node-id=2115%3A371
- removed preferences page entirely
- removed tags for study environment (solo/collab) and meeting type (online/in-person). removed associated variables in StudyGroup struct. resized cards/cells to match the removal of these tags
- modified Get Started button (when user isn't signed in) so the text and button color match Figma
- added Study Pact label at the top of the view showing all study groups
- added "Pending" page for when user taps a cell with a pending study group
  - remove button brings up an alert asking for confirmation
- changed "Create Group" to "Create Preference" to match Figma
- reorganized the StudyPact colors

screenshots:
IMG_1934
IMG_1935
IMG_1936
IMG_1937
IMG_1939


Stereopoly:
A few comments/thoughts
* The new salmon colored sign-in button needs a darker shade for the .pressed mode
* I think having the "Remove Pending Request" as grey makes it seem like it's a disabled button, maybe the Blue button could work here
* For the groups, there should be a check on the size of the group, e.g. having 10 members would probably result in super-overlapped images

Stereopoly (approved):

KevWho:
+1 on Oscar's comment regarding the .pressed differences. We could also add some response for users tapping on study group cells and the "Create Preference" button, similar to the response when users tap on Libraries, Dining Halls, etc.

Would we want to display the Study Pact page in a drawer (like when a Library is opened) instead of a modal?

Also, extra folder in the project file:
Screen Shot 2021-01-30 at 4 09 36 PM

huangshawn1:
color changes will be made once design decides on the colors. don't think we want to present the page as a drawer and it's a bit more complicated to do as well. I don't see the extra folder on my end.

PR #248, opened 2021-01-31
Eashan study pact onboarding

eashan-mathur:


berkeley-mobile/StudyPact/PageViewController.swift
```diff
@@ -0,0 +1,219 @@
...
+//  PageViewController.swift
+//  berkeley-mobile
+//
+//  Created by Eashan Mathur on 10/17/20.
+//  Copyright © 2020 ASUC OCTO. All rights reserved.
+//
+
+import UIKit
+
+class PageViewController: UIPageViewController, UIPageViewControllerDataSource, UIPageViewControllerDelegate {
```
KevWho, line 11: Is this file a copy of `/StudPact/Onboarding/PageViewController.swift`? Doesn't seem to be used in the project rn.

berkeley-mobile/StudyPact/CreatePreference/SelectLocationView.swift
```diff
@@ -0,0 +1,102 @@
...
+            sender.isSelected.toggle()
+            if let buttonSelected = self.buttonSelected {
+                buttonSelected.backgroundColor = .white
+                buttonSelected.setTitleColor(Color.blackText, for: .normal)
+                sender.backgroundColor = Color.ActionButton.background
+                sender.setTitleColor(.white, for: .normal)
+            }
+            self.buttonSelected = sender
+        }
+    }
```
KevWho, line 64: If `buttonSelected.isSelected` is never set to false when another option is chosen, so choosing "Virtual", then "In Person", then attempting to re-select "Virtual" doesn't do anything.

KevWho (changes requested):
Needed to do a `pod install` to get the new dependency, might want to add those changes to this PR too, since we've pushed all our other pods. Also, would we want to remove these unused / old onboarding images?

Screen Shot 2021-02-02 at 3 20 43 PM

berkeley-mobile/StudyPact/CreatePreference/SelectClassView.swift
```diff
@@ -0,0 +1,95 @@
...
+import UIKit
+import SearchTextField
+
+class SelectClassView: UIView, UITextFieldDelegate, EnableNextDelegate {
+    let preferenceVC: CreatePreferenceViewController
+    var textField: SearchTextField!
+    var classNames: [String] = []
+    override func layoutSubviews() {
+        super.layoutSubviews()
+    }
```
KevWho, line 18: Doesn't seem to be needed

berkeley-mobile/StudyPact/CreatePreference/SelectClassView.swift
```diff
@@ -0,0 +1,95 @@
...
+//
+//  Created by Eashan Mathur on 1/30/21.
+//  Copyright © 2021 ASUC OCTO. All rights reserved.
+//
+
+import UIKit
+import SearchTextField
+
+class SelectClassView: UIView, UITextFieldDelegate, EnableNextDelegate {
+    let preferenceVC: CreatePreferenceViewController
```
KevWho, line 13: Someone correct me if I'm wrong on this, but this would create a retain cycle (`SelectClassView` has a strong reference to `preferenceVC`, which has a strong reference to `SelectClassView` through `preferenceVC.pages`). Could fix by making this `weak` or something similar.

Applies to `SelectClassView` and the other classes with strong references to `CreatePreferenceViewController`.

KevWho (changes requested):
Sorry, a couple more small things

KevWho (approved):
Thanks!

PR #251, opened 2021-02-03
Onboarding make launch screen

eashan-mathur:


berkeley-mobile/MainContainerViewController.swift
```diff
@@ -47,6 +49,17 @@ class MainContainerViewController: UIViewController, MainDrawerViewDelegate {
...
+        
+        let launchedBefore = UserDefaults.standard.bool(forKey: "launchedBefore")
+        if launchedBefore  {
+           print("Not first launch")
+        } else {
+            let vc = PageViewController(transitionStyle: .scroll, navigationOrientation: .horizontal)
+            vc.modalPresentationStyle = .fullScreen
+            self.present(vc, animated: true, completion: nil)
+            print("First launch")
+            UserDefaults.standard.set(true, forKey: "launchedBefore")
```
huangshawn1, line 61: change the name of the key to something like "hasShownStudyPact" because we may want to use this sort of thing for future feature onboardings. "launchedBefore" is a bit ambiguous/general

huangshawn1: also you can remove the prints

huangshawn1 (approved):

PR #254, opened 2021-02-03
API Calls framework/base

huangshawn1:
Add StudyPact class to hold all functions for api calls, user cryptohash and email.
Left TODO markers for where the backend api should be called in various places.
Implemented RegisterUser and GetGroups api calls (don't know if they work because of backend issues).


berkeley-mobile/Libraries/StudyViewController.swift
```diff
@@ -128,50 +128,25 @@ class StudyViewController: UIViewController, UITableViewDataSource, UITableViewD
...
-    
     /// modify the contents of the card based on if the user is signed in, what groups the user is in
     func refreshStudyGroupContents() {
-        // TODO: once auth/profile is done set to use actual value. using time to test both forms
-        let date = Date()
-        let calendar = Calendar.current
-        let seconds = calendar.component(.second, from: date)
-        let loggedIn = seconds % 2 == 1
-        
+        let loggedIn = true//StudyPact.shared.getCryptoHash() != nil
```
KevWho, line 133: I'm assuming we're changing this soon haha

KevWho (approved):
I think I recently changed `NetworkManager` to call completion on the main thread, so the calls to `DispatchQueue.main.async` in `StudyGroupsView`, etc. shouldn't be necessary. Don't quote me on that though.

berkeley-mobile/StudyPact/StudyPact.swift
```diff
@@ -0,0 +1,151 @@
...
+import Foundation
+import GoogleSignIn
+
+/// All StudyPact api call functions go in here. Relevant user info also goes here.
+class StudyPact {
+    static let shared = StudyPact()
+    
+    private var cryptoHash: String?
+    private var email: String?
+    var studyGroups: [StudyGroup] = []
```
KevWho, line 18: What is this used for?

huangshawn1: this is for storing the user's info so that it's centralized

KevWho: For `studyGroups` specifically though, it doesn't seem to be used for any purpose currently, and wouldn't we fetch groups from the API instead of storing it here?

huangshawn1: ah ok yeah I removed studygroups now

huangshawn1:
Added a SignInManager that handles all the google sign in stuff. This was to resolve a couple issues.

StudyPact.shared.email needed to be updated as soon as a previous sign in was restored, but this can only be done in a GIDSignInDelegate. Doing it from the ProfileViewController (the previous delegate) didn't make sense because then it would only happen if the user went to the Profile page.

Moving to SignInManager allows the email to be updated on startup. Allows multiple delegates for future by adding an array of delegates that it manages.
Also, the bit of lag on the Profile Page when it has to restore the previous sign (when the app is launched again but the user is signed into google) is gone because the sign in is restored on launch.

berkeley-mobile/StudyPact/Profile/ProfileViewController.swift
```diff
@@ -38,15 +38,13 @@ class ProfileViewController: UIViewController, UITextFieldDelegate, UINavigation
...
         
-        GIDSignIn.sharedInstance()?.delegate = self
-        let previousSignIn = GIDSignIn.sharedInstance().hasPreviousSignIn()
+        SignInManager.shared.addDelegate(delegate: self)
         
-        if let cryptoHash = StudyPact.shared.getCryptoHash(), previousSignIn {
-            // TODO: call GetUser and fill in profile
-            GIDSignIn.sharedInstance()?.restorePreviousSignIn()
+        if SignInManager.shared.isSignedIn,
+           let cryptohash = StudyPact.shared.getCryptoHash() {
```
KevWho, line 44: Isn't the `let cryptohash = StudyPact.shared.getCryptoHash()` line redundant since `.isSignedIn` checks all that jazz?

berkeley-mobile/StudyPact/Profile/ProfileViewController.swift
```diff
@@ -60,35 +58,18 @@ class ProfileViewController: UIViewController, UITextFieldDelegate, UINavigation
...
-                    switch result {
-                    case .success(let image):
-                        DispatchQueue.main.async() { [weak self] in
-                            self!.profileImage = image
-                        }
-                    case .failure(let error):
-                        print(error)
+        // if new sign in to google rather than previous sign in
+        if StudyPact.shared.getCryptoHash() == nil {
+            StudyPact.shared.registerUser(user: user) { success in
```
KevWho, line 63: What do you think about moving this block to `SignInManager` before this method is called on the delegates, so all the authentication is centralized there?

berkeley-mobile/StudyPact/StudyPact.swift
```diff
@@ -0,0 +1,150 @@
...
+class StudyPact {
+    static let shared = StudyPact()
+    
+    private var cryptoHash: String?
+    var email: String?
+    
+    /// Load cryptohash from user defaults. If it doesnt exist or authenticate fails, return false.
+    public func loadCryptoHash() -> Bool {
+        reset()
+        if let cryptoHash = UserDefaults.standard.string(forKey: "userCryptoHash") {
```
KevWho, line 22: `reset()` deletes the CryptoHash, so this if statement is never true and the user must re-authenticate each time the app is started.

huangshawn1: oh yeah lmao that was for testing

berkeley-mobile/StudyPact/Profile/ProfileViewController.swift
huangshawn1: this was just to unwrap the optional. could force unwrap I guess but I think this is better

huangshawn1: this was to unwrap it because the cryptohash was gonna be used here, but I ended up removing the thing that was gonna use the cryptohash

huangshawn1: removed this now

berkeley-mobile/StudyPact/Profile/ProfileViewController.swift
huangshawn1: so essentially this is if the user wasn't previously signed in and has just signed in using the "Sign in with Google" button. so this should theoretically only happen on the Profile page. the check that the cryptohash is nil is essentially checking for this.
we want this because we only want to register the user if the cryptohash expired/doesn't exist and the user just signed into google

PR #258, opened 2021-02-06
Implement Authenticate and Class Preference API

KevWho:
Add call to `StudyPact.authenticate` during `StudyPact.loadCryptoHash`. Add `.cancelPending` and hook up `.addClass` and `.cancelPending` to their locations in the UI.
- Add failure alerts when adding / removing pending class preferences fails.
- Rename `StudyPact/API/Schema/AddClass.swift` > `ClassPreference.swift`
  - Rename `AddClassParams` > `AddPreferenceParams`
  - Add `RemovePreferenceParams`
- Add `size` and `id` properties to `StudyGroup`
- Add print statements on network request failure to aid in future debugging.


huangshawn1 (approved):
looks good to me (assuming cancel pending and add class work once get groups is in place)

PR #260, opened 2021-02-10
Study pact wrapup

huangshawn1:
- Fix bug where tapping home button to collapse a drawer would cause a crash after another drawer had been dismissed
- Add GroupUpdateDelegate to allow StudyGroupsView to refresh when AddClass, CancelPending, etc is called successfully
- Switch GetGroups to use GET instead of POST so it works now
- Allow demo email to log in even though it isn't @berkeley.edu email (contained in Secrets file)
- Add placeholder for profile picture in GroupMembersView
- Add error message on register failure
- Remove self from groups


Stereopoly:
The functionality all seems to work well apart from leaving a group which gave me an error message.

Stereopoly (approved):

PR #262, opened 2021-02-13
Study Pact Hotfixes

KevWho:
- Specific error message for when user signs in with a non-berkeley email (right now it’s a general failed to register)
- Group members view: use blue/grey icons to indicate availability, disable button instead of showing a popup
- Go to main study page with drawer up after user is added for the first time
- New, more descriptive onboarding screens (see Figma)
- Logout bug
- Bold “your-username” for profile facebook tooltip


Stereopoly:
If user selects "Sign up later." they will never see the onboarding flow. Maybe we either not flag the launch and show the onboarding next time, or show it when they navigate to the profile page.

Stereopoly (approved):

PR #264, opened 2021-02-14
Leave Group and Add Class changes

huangshawn1:
Leave group now shows a popup to confirm and dismisses on success. The backend works as well.
Add class has been changed to send preferred number of people to study with + 1 because backend wants the total size of the group (including self).


KevWho (approved):
Code looks good. Works on the demo account, had some issues with backend's response on my @berkeley.edu account.

PR #266, opened 2021-02-16
Change location option to environment

huangshawn1:
Instead of in-person vs virtual, we switched to collaborative vs quiet environment.
Also updated the review string to say "person" for size 1.


Stereopoly:
Looks good!

Stereopoly (approved):

PR #267, opened 2021-02-22
Kevinwang/mapmarkerswipe

Kwang8:
Added a swipe down gesture that calls closeview function.

Added a UIViewTransition to closeview that would show a flipping animation to close the app. Commented out delegate?.didCloseMarkerDetailView(self) as it interfered with the animation, don't know if this is ok or not.

### DIFF
--- a/berkeley-mobile/Map/MapMarkerDetailView.swift
+++ b/berkeley-mobile/Map/MapMarkerDetailView.swift
@@ -31,9 +31,9 @@ class MapMarkerDetailView: UIView {
                     self.alpha = 1.0
                 }) { completed in self.isHidden = !completed }
             } else {
-                UIView.animate(withDuration: 0.2, animations: {
-                    self.alpha = 0.0
-                }) { completed in  self.isHidden = completed }
+                UIView.transition(with: self, duration: 0.5, options: [.transitionFlipFromBottom], animations: {
+                    self.alpha = 0
+                }, completion: nil)
             }
         }
     }
@@ -120,7 +120,6 @@ class MapMarkerDetailView: UIView {
     
         verticalStack.bottomAnchor.constraint(equalTo: contentView.layoutMarginsGuide.bottomAnchor).isActive = true
         
-        //swipe down gesutre
         let swipeDown = UISwipeGestureRecognizer(target: self, action: #selector(closeView))
         swipeDown.direction = .down
         self.addGestureRecognizer(swipeDown)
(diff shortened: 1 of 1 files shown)


Stereopoly (commented):
There doesn't seem to be much of a flipping animation for me - It begins but then disappears super fast. 

Also probably unrelated to this, but we have a feature that dismisses any view if you tap the home button when already on the home page, but it doesn't dismiss the MapMarkerDetailViews.

berkeley-mobile/Map/MapMarkerDetailView.swift
```diff
@@ -120,6 +119,12 @@ class MapMarkerDetailView: UIView {
         detailStack.alignment = .center
     
         verticalStack.bottomAnchor.constraint(equalTo: contentView.layoutMarginsGuide.bottomAnchor).isActive = true
+        
+        //swipe down gesutre
```
huangshawn1, line 123: typo, also you don't really need to comment this because the code is pretty self-explanatory here

berkeley-mobile/Map/MapMarkerDetailView.swift
```diff
@@ -168,10 +173,16 @@ class MapMarkerDetailView: UIView {
...
     @objc func closeView() {
+        
         marker = nil
-        delegate?.didCloseMarkerDetailView(self)
+        UIView.transition(with: self,
+                          duration: 1.0,
+                                 options: [.transitionFlipFromBottom],
+                                 animations: {
+                                 },
+                                 completion: nil)
```
huangshawn1, line 183: the spacing/indent is a bit off here, you cant hit ctrl+i to auto indent or adjust yourself

berkeley-mobile/Map/MapMarkerDetailView.swift
```diff
@@ -168,10 +173,16 @@ class MapMarkerDetailView: UIView {
...
     }
-
+    //self.delegate?.didCloseMarkerDetailView(self)
```
huangshawn1, line 185: instead of commenting this out, you should add the animation to these lines: https://github.com/asuc-octo/berkeley-mobile-ios/blob/cae66f01b93753c2133770086fcf909c95f4e3a1/berkeley-mobile/Map/MapMarkerDetailView.swift#L34-L36
this is the code that is run when the detail is dismissed.

huangshawn1 (changes requested):

huangshawn1 (approved):
looks good to me

PR #269, opened 2021-03-06
Make pins bigger when selected.

Kwang8:
Added a view.tranform to make annotations 1.5 times bigger when selected then back to original when deselected.


huangshawn1 (commented):
looks good, is there a way to add an animation so it slowly grows? right now it looks strange when it suddenly increases in size

Kwang8:
Added a UIView.animate so the pins would get better with a .1 second animation.

huangshawn1 (approved):
looks good

PR #270, opened 2021-03-06
Map Marker Collection

KevWho:
Changes map markers to pull data from the `"Map Marker"` collection in Firebase instead of individual collections for each category. Previously, a hardcoded set of categories (Mental Health, Water, etc.) are always shown on the map. Changes this so that the buttons are based on marker data from Firebase (i.e. empty / depricated categories are not shown, and unrecognized / new categories are still shown).
- Updates `MapMarkerType`:
  - Adds `.store`, `.restaurant` cases and associated color / icon for `.store`
  - Removes `.mentalHealth` since this is to be moved to the resources tab
  - Adds `.none` case for color / icon for any other markers with types not enumerated
- Updates `MapMarker` model with additional (but currently unused) fields from Firebase
- Changes `MapMarker.type` to be either `.known` with an associated `MapMarkerType` or `.unknown` with some arbitrary String value.
- Moves creation of `MapViewController.filters` to after map markers have been fetched / parsed.
  - Known categories will be listed first, in the order they were declared in `MapMarkerType`, then unrecognized categories after in an arbitrary order
  - Change `FilterView` slightly to animate in the change


huangshawn1 (commented):
add mental health icon back for now until it's moved. store icon looks a bit low quality compared to the other icons. placeholder icon also looks a bit fuzzy.

huangshawn1 (approved):
looks good

PR #272, opened 2021-03-07
AlertVC

eashan-mathur:
image


Stereopoly:
Really like the design! Any way we could move the buttons a little closer?
image

eashan-mathur:
Fixed the spacing issue Oscar addressed above

berkeley-mobile/Common/AlertView.swift
```diff
@@ -0,0 +1,161 @@
...
+        btn2.addTarget(self, action: #selector(btn2Function), for: .touchUpInside)
+        
+        btn1.titleLabel?.adjustsFontSizeToFitWidth = true
+        btn2.titleLabel?.adjustsFontSizeToFitWidth = true
+        
+        btn1.layer.shadowOpacity = 0
+        btn2.layer.shadowOpacity = 0
+        
+        
+
```
huangshawn1, line 73: should probably cut down on the amount of whitespace here

berkeley-mobile/Common/AlertView.swift
```diff
@@ -0,0 +1,161 @@
...
+        Hstack.spacing = 35
+        
+        Hstack.addArrangedSubview(btn1)
+        Hstack.addArrangedSubview(btn2)
+        Hstack.translatesAutoresizingMaskIntoConstraints = false
+        
+        backgroundView.addSubview(Vstack)
+        backgroundView.addSubview(Hstack)
+        
+        print("setting up")
```
huangshawn1, line 139: remove print

berkeley-mobile/StudyPact/Profile/ProfileViewController.swift
```diff
@@ -186,9 +187,26 @@ extension ProfileViewController {
...
+//            StudyPact.shared.reset()
+//
+//            self.loggedOutView()
+//        }))
+//
+//        self.present(alertController, animated: true, completion: nil)
+        
+        let alertView = AlertView(headingText: "Logout?", messageText: "You will be returned to the login screen.", action1Label: "Cancel", action1Color: Color.ActionButton.background, action1Completion: {
+            self.dismiss(animated: true, completion: nil)
+        }, action2Label: "Logout", action2Color: .gray, action2Completion: {
```
huangshawn1, line 209: probably don't want to use .gray, instead add a new color to Color

huangshawn1 (changes requested):
remove the pod because it's never used.
generally looks really good. I think you can start replacing existing UIAlertControllers throughout the rest of the app as well. some of them have a single button though so maybe generalize to account for that

objorkman:
There seem to be some width constraint issues:
Simulator Screen Shot - iPhone 12 Pro - 2021-04-10 at 16 13 53
Simulator Screen Shot - iPhone 12 Pro - 2021-04-10 at 16 15 37

objorkman:
Learn More and Register use the old Alert

Simulator Screen Shot - iPhone 12 Pro - 2021-04-10 at 16 29 37

objorkman:
And the address
Simulator Screen Shot - iPhone 12 Pro - 2021-04-10 at 16 30 40

berkeley-mobile/Common/AlertView.swift
```diff
@@ -0,0 +1,190 @@
...
+        }
+        Hstack.translatesAutoresizingMaskIntoConstraints = false
+        
+        backgroundView.addSubview(Vstack)
+        backgroundView.addSubview(Hstack)
+        
+        
+        backgroundView.centerXAnchor.constraint(equalTo: view.centerXAnchor).isActive = true
+        backgroundView.centerYAnchor.constraint(equalTo: view.centerYAnchor).isActive = true
+        backgroundView.widthAnchor.constraint(equalTo: view.widthAnchor, multiplier: 304/375).isActive = true
```
huangshawn1, line 167: nitpick: reduce whitespace, we typically don't want to have 2 lines of whitespace in a row

berkeley-mobile/Data/ItemProtocols/HasOpenTimes.swift
```diff
@@ -20,20 +20,20 @@ extension HasOpenTimes {
...
+                        let interval = DateInterval(start: openDate, end: closeDate)
+                        weeklyHours.addInterval(interval, to: DayOfWeek.weekday(openDate))
+                    }
                 }
-                if openDate <= closeDate {
-                    let interval = DateInterval(start: openDate, end: closeDate)
-                    weeklyHours.addInterval(interval, to: DayOfWeek.weekday(openDate))
-                }
-            }
         }
```
huangshawn1, line 37: I think this change isn't supposed to be here? seems unrelated

berkeley-mobile/StudyPact/Profile/ProfileViewController.swift
```diff
@@ -186,9 +186,11 @@ extension ProfileViewController {
...
     
     @objc private func logoutButtonPressed(sender: UIButton) {
-        let alertController = UIAlertController(title: "Logout?", message: "You will be returned to the login screen.", preferredStyle: .alert)
-        alertController.addAction(UIAlertAction(title: "Cancel", style: UIAlertAction.Style.cancel, handler: nil))
-        alertController.addAction(UIAlertAction(title: "Logout", style: UIAlertAction.Style.destructive, handler: {_ in
+        
+        
+        let alertView = AlertView(headingText: "Logout?", messageText: "You will be returned to the login screen.", action1Label: "Cancel", action1Color: Color.ActionButton.background, action1Completion: {
+            self.dismiss(animated: true, completion: nil)
+        }, action2Label: "Logout", action2Color: Color.AlertView.secondaryButton, action2Completion: {
```
huangshawn1, line 193: looks like the cancel button is blue here, whereas in other places the cancel button is gray.
I think we should make the cancel button gray here and keep that pretty consistent

berkeley-mobile/Utils/UIViewController+Extensions.swift
```diff
@@ -28,22 +28,34 @@ extension UIViewController {
     
     // Presents an alert with multiple options and completion handler
     public func presentAlertLinkUrl(title: String, message: String, options: String..., website_url: URL) {
-        let alertController = UIAlertController(title: title, message: message, preferredStyle: .alert)
+        
+        var alertVC: AlertView!
+
```
huangshawn1, line 33: I think this function is meant to present any number of options and it loops through all of them, adding them to the alert.
Because your AlertView doesn't support this (and we don't use that in the app), you should probably change this function to only take in title, message, and a single button text option.
Then, get rid of the loop and just present your AlertView with the given title, message, a cancel button, and the other button that opens the url.

berkeley-mobile/Utils/UIViewController+Extensions.swift
```diff
@@ -28,22 +28,34 @@ extension UIViewController {
...
-                alertController.addAction(UIAlertAction.init(title: option, style: .cancel))
+                alertVC = AlertView(headingText: title, messageText: message, action1Label: option, action1Color: Color.AlertView.secondaryButton, action1Completion: {
+                    self.dismiss(animated: true, completion: nil)
+                }, action2Label: "", action2Color: UIColor.clear, action2Completion: {return}, withOnlyOneAction: true)
             } else {
-                alertController.addAction(UIAlertAction.init(title: option, style: .default, handler: { _ in
+                
+                alertVC = AlertView(headingText: title, messageText: message, action1Label: "Cancel", action1Color: Color.AlertView.secondaryButton, action1Completion: {
+                    self.dismiss(animated: true, completion: nil)
+                }, action2Label: "Yes", action2Color: Color.ActionButton.background, action2Completion: {
```
huangshawn1, line 43: this is a bug/issue from before, but we should dismiss the alert after opening the url.
right now if the user hits "Yes", the url opens but when they go back to the app the AlertView is still there

berkeley-mobile/Utils/UIViewController+Extensions.swift
```diff
@@ -28,22 +28,34 @@ extension UIViewController {
...
+        alertVC.modalPresentationStyle = .overCurrentContext
+        
+        self.present(alertVC, animated: true, completion: nil)
     }
     
     // Presents an alert to open coordinates
     public func presentAlertLinkMaps(title: String, message: String, options: String..., lat: CLLocationDegrees, lon: CLLocationDegrees, name: String) {
-        let alertController = UIAlertController(title: title, message: message, preferredStyle: .alert)
+        
+        var alertVC: AlertView!
```
huangshawn1, line 58: same 2 changes as presentAlertLinkUrl should be applied to presentAlertLinkMaps

berkeley-mobile/Data/ItemProtocols/CalendarEvent.swift
```diff
@@ -39,19 +39,25 @@ extension CalendarEvent {
...
                     } else {
+                        vc.dismiss(animated: true, completion: nil)
                         vc.presentFailureAlert(title: "Failed to add to calendar", message: "Make sure Berkeley Mobile has access to your calendar and try again.")
                     }
                 }
             }
-        }))
+        }, withOnlyOneAction: false)
+        alertController.modalPresentationStyle = .overCurrentContext
+        alertController.modalTransitionStyle = .crossDissolve
```
huangshawn1, line 60: it looks like these two lines are used every time so you can just put them in the AlertView class. for example, just setting `self.modalPresentationStyle = .overCurrentContext` and `self.modalTransitionStyle = .crossDissolve` in the init

huangshawn1 (changes requested):
looks really good, a few changes I think we should make before merging

berkeley-mobile/Utils/UIViewController+Extensions.swift
huangshawn1: the function here is`presentAlertLinkUrl`

huangshawn1 (approved):
looks great, good work

PR #273, opened 2021-03-13
Re-Add Covid Page

Stereopoly:
* Thoughts on statistics headers? Too "covid related"?
image


huangshawn1:
> * Thoughts on statistics headers? Too "covid related"?
figma has "total" and "new" so  maybe "total" and "percentage"/"rate"

huangshawn1 (approved):

PR #275, opened 2021-03-14
Mapmarker image change

Stereopoly:
Updates pdfs to use svgs instead, except for Waste icon. Fixes lag in switching to Rest markers


huangshawn1 (approved):
looks good to me but I'll let @KevWho look at it and merge cause its his branch

PR #276, opened 2021-04-03
Property Wrapper for Display Strings

KevWho:
Cleans up invalid characters or extra whitespace in some display strings pulled from Firebase.
- Adds `Display` property wrapper
  - Trims whitespace, removes `�` character
- Wraps important fields for Libraries, Gyms, Dining, Campus Resources, Events, etc.
  - `name`, `address` / `location`, `description`, `email`, `phone`, ...


huangshawn1 (commented):
looks good. some other place we might want to add the display wrapper:
CovidResource
GymClass.trainer (this isn't actually shown in-app but is used to create the calendar event when you add a gym class to the calendar). it looks like when I add Yoga Stretch to my calendar the trainer is "`&nbsp;`Michael", so is there a way to filter that out as well?

KevWho:
Added to `CovidResource` and `GymClass.trainer`.

> it looks like when I add Yoga Stretch to my calendar the trainer is &nbsp;Michael, so is there a way to filter that out as well?

Hmm, there's a way to parse HTML symbols into their display strings (e.g. `&lt;` -> `<`) using `NSAttributedString`. I tried it, and it looks like the method is pretty expensive, so it's really slow when used on all the display strings for all the resources. What do you think?

huangshawn1:
> Hmm, there's a way to parse HTML symbols into their display strings (e.g. `&lt;` -> `<`, `&nbsp;` -> ` `, etc.) using `NSAttributedString`. I tried it, and it looks like the method is pretty expensive, so it's really slow when used on all the display strings for all the resources. What do you think?

that's fine then, not a big enough improvement to justify making everything slower

huangshawn1 (approved):

PR #277, opened 2021-04-03
Show Mental Health Resources under Health Tab

KevWho:
Adds `"Mental Health"` to the list of known resource types / colors, set up the `"Health"` tab to contain both health and mental health resources.


huangshawn1 (approved):
looks good

PR #283, opened 2021-09-06
added search functionality

Kwang8:
Made map markers searchable. @KevWho helped a lot.


objorkman:
Some issues with duplicate results, an example is when searching for Marugame Udon

huangshawn1:
to address the duplicate issue: we decided to remove cafes entirely so that they don't conflict with the map markers in search. currently cafes only show up in search so no other part of the app will be affected. @Kwang8 could you go ahead and delete Cafe, CafeDataSource, CafeItem, and all other related files/code?

berkeley-mobile/Map/MapViewController.swift
```diff
@@ -402,6 +402,7 @@ extension MapViewController: SearchResultsViewDelegate {
     // drop new pin and show detail view on search
     func choosePlacemark(_ placemark: MapPlacemark) {
         // remove last search pin
+        removeAnnotations(type: MapMarker.self)
```
huangshawn1, line 405: this leads to the bug where if all Cafe map markers are visible for example, dropping any pin/map marker using search will cause the Cafe map markers to disappear.
I like your solution where we track the single current pin/map marker dropped using search. then we can remove only that annotation instead of removing all annotations.

berkeley-mobile/Map/MapViewController.swift
```diff
@@ -400,10 +403,20 @@ extension MapViewController: SearchBarDelegate {
 extension MapViewController: SearchResultsViewDelegate {
     
     // drop new pin and show detail view on search
+    //var previousMapMarker: MapMarker?
+    //var previousPlaceMark: MapPlacemark?
```
huangshawn1, line 407: remove

berkeley-mobile/Map/MapViewController.swift
```diff
@@ -400,10 +403,20 @@ extension MapViewController: SearchBarDelegate {
...
     func choosePlacemark(_ placemark: MapPlacemark) {
         // remove last search pin
-        removeAnnotations(type: MapMarker.self)
-        removeAnnotations(type: SearchAnnotation.self)
+        
+        if((self.previousPlaceMark) != nil) {
```
huangshawn1, line 411: we prefer to use something like
`if let previousPlaceMark = self.previousPlaceMark`
here instead of checking nil then force unwrapping

huangshawn1 (commented):
still one small issue: if 1 show all cafe map markers, then search marugame, then search something else like Moffitt, the cafe map marker for marugame will disappear

berkeley-mobile/Map/MapViewController.swift
```diff
@@ -437,6 +452,38 @@ extension MapViewController: SearchResultsViewDelegate {
...
+        let location = mapMarker.location
+        if location.0 != Double.nan && location.1 != Double.nan {
+            let regionRadius: CLLocationDistance = 250
+            // center map on searched location
+            let coordinateLocation:CLLocationCoordinate2D = CLLocationCoordinate2DMake(location.0, location.1)
+            let coordinateRegion = MKCoordinateRegion(center: coordinateLocation,
+                                                      latitudinalMeters: regionRadius * 2.0, longitudinalMeters: regionRadius * 2.0)
+            mapView.setRegion(coordinateRegion, animated: true)
+            mapView.addAnnotation(mapMarker)
+            self.previousMapMarker = mapMarker
```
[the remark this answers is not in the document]
Kwang8: It's still adding the marugame annotation for me even when I selected cafe markers. @huangshawn1

huangshawn1 (approved):
looks good

PR #285, opened 2021-09-27
Drop pin and set detail to middle position

huangshawn1:
Drop pin and set detail to middle position when opening from table.
Table tap behavior and search behavior should match now.

do we want to set an offset to make it obvious that the detail drawers can be pulled to the full position? right now it might not look like they are draggable


objorkman (approved):
lgtm!

PR #287, opened 2021-10-07
Fix bug when no classes available

objorkman:
Fix issue that scroll view extends past content height when there are no classes available


huangshawn1 (approved):
lgtm

PR #289, opened 2021-10-14
Sproul.club Card

KevWho:
Embeds the sproul.cub webview inside a card with rounded corners and a drop shadow to make it appear less out-of-place.


huangshawn1 (approved):
lgtm

PR #291, opened 2021-10-14
Dining Menu Swiping

KevWho:
Adds ability to swipe left/right through dining menus.
- Moves code for the `FilterTableView` into a new `DiningMenuViewController`
- Replaces the `SegmentedControl` with a `SegmentedControlViewController` with pages of `DiningMenuViewController`

To make the dietary restriction filter selections shared between menus:
- Adds "header" section to `SegmentedControlViewController`
- Adds option for `FilterTableView` to use a pre-existing `FilterView` instead of creating its own
- Marks each menu table as a delegate of the single `FilterView`
  - Would probably be better to use Notification Center, but this would involve some more changes.


huangshawn1 (approved):
looks good

PR #296, opened 2021-11-04
Dining Hall Scroll Bug

KevWho:
Fixes the issue where opening a detail view (e.g. clicking on a Library), then selecting a Dining Hall yields weird behavior when attempting to move the drawer.


huangshawn1 (approved):
looks good

PR #298, opened 2021-11-04
Fix detail views

Kwang8:


huangshawn1 (approved):
a is for awesome - Kevin Hu

PR #299, opened 2022-03-12
Sort / Select Dining Hall Menu by Current Time

KevWho:
## Overview

Sorts Dining Hall menus by the open/close hours of each menu, if that data is available (e.g. "Brunch" generally comes before "Dinner"). Additionally, automatically selects the menu corresponding to the meal taking place at the current time, if any (e.g. "Dinner" tab will be opened initially around dinner time).

## Description

- Parse "note" field from Firebase for each location
  - For dining halls, this is the meal taking place at that interval
  - Adds `HoursInterval` when parsing weekly hours, which stores the note in addition to the date interval
- If all dining hall menus have a corresponding open/close interval:
  - Sorts the menus by these date intervals
  - Sets the selected menu index to the one happening right now (or 0)


huangshawn1:
everything looks good, but we should automatically go to the next meal if the current time isn't in any interval

huangshawn1 (approved):
lgtm

PR #301, opened 2022-10-08
UI Change for Study Group Search Request Button

Sukhamrit-Singh:
Changed the color of the button since it looked like it was not enabled before. Switched it from a gray to red color. Created another color in the color files just for this button.

Before UI Change:
IMG_1919

After UI Cha
Simulator Screen Shot - iPhone 14 Pro - 2022-10-07 at 17 07 13
nge:


huangshawn1 (approved):
lgtm

PR #303, opened 2022-10-08
Profile Page Facebook Username Example Added

chloeyeoo:
1. changed default text below facebook-username field from your-username to oski-bear
2. added oski-bear as default example into facebook username field
3. had to redownload the GoogleServiceInfo.plist (file from onboarding day) because it wasn't building on simulator, but it works now :)


objorkman (approved):
LGTM!

PR #304, opened 2022-11-19
added safety to top scroll bar and deleted other options

sydneytsai:
- added color values for safety in Colors+Resource file
- added Safety to ResourceType file
- added Safety to ResourcesViewController file under setupSegementedControls
- deleted dashboard, clubs, and other from the scroll bar
- rearranged the scroll bar to safety, health, basic needs, admin


huangshawn1 (changes requested):
we can make a new SafetyResourceViewController or similar instead of using the CampusResourceViewController because the safety tab doesn't quite follow that same format. try making something similar to the CovidResourceViewController

sydneytsai:
- deleted safety from resourcesviewcontroller and color
- added a safetyresourcesviewcontroller file (based on the covid one) and linked that to the top bar instead of using resourcesviewcontroller

huangshawn1 (approved):
lgtm

PR #306, opened 2024-03-05
Finish GIRs, Basic Resources Tab, Clean Up Project Files

justinwongeecs:
Finish gender inclusive restrooms feature & implemented the general resources tab feature at its basics. 
Resources Tab fetches resource categories and shoutouts from Firestore. 

Cleaned up project files:
- Deleted Covid Dashboard and old resources section 
- Updated Pod files so that we can use Xcode 15 (previously the project won't work in Xcode 15 since some dependencies were so out of date)


Sukhamrit-Singh:
looks good

PR #316, opened 2024-11-13
BER-9: Feedback Form Enhancements

justinwongeecs:
**Feedback Form Enhancements:**
Linear Link: https://linear.app/moon-stone/issue/BER-9/feedback-form-enhancements  
- Only appears after initial app use (2nd launch)
- Option to dismiss anytime with close button in top trailing toolbar. Upon doing so, feedback form will not be shown again. 
- Submit button has 3D effect view 
- Switch "Other Comments" Section text input view to TextEditor

https://github.com/user-attachments/assets/d7a5c77e-959a-489c-a685-88de078fb34a


sahanabhar (approved):
These changes look great - will definitely enhance user accessibility with the feedback form!

PR #323, opened 2025-02-21
BER-49: Improve SegmentedControlView's Indicator Bar

baeuke:
- Moved the indicator down and changed its height (thickness) so it didn't occlude with the tab names
- Experimented with the numbers for height and spacing. 
   - weird behavior happened for indicatorOffset of -30 and indicatorHeight of 8: the tabs were not tappable in the preview (exaclty for these numbers); 
   - and for some other numbers the tabs would only be tapped and selected if tapped on the center of the text (tab name) and not the edges
- small refactoring helped to make the tapping behavior more deterministic
   - reordered the `Text` and `capsule` views in the hierarchy and adjusted the numbers
   - also added `label` for the button (for readability)
<div align="center">
Screenshot 2025-02-20 at 10 12 52 PM
</div>

### DIFF
--- a/berkeley-mobile.xcodeproj/project.pbxproj
+++ b/berkeley-mobile.xcodeproj/project.pbxproj
@@ -158,13 +158,13 @@
 		55DCF7A123724835001B01B8 /* MaterialButton.swift in Sources */ = {isa = PBXBuildFile; fileRef = 55DCF7A023724835001B01B8 /* MaterialButton.swift */; };
 		E80330EB2CE431C200DC9574 /* DepthButtonStyle.swift in Sources */ = {isa = PBXBuildFile; fileRef = E80330EA2CE431C200DC9574 /* DepthButtonStyle.swift */; };
 		E80330ED2CE7EBB100DC9574 /* NSCoding+Extension.swift in Sources */ = {isa = PBXBuildFile; fileRef = E80330EC2CE7EBB100DC9574 /* NSCoding+Extension.swift */; };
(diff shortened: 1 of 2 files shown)


justinwongeecs:
Why are there changes to the podfiles and to the `project.pbxproj` file? The only changes should be in `SegmentedControlView.swift`.

berkeley-mobile/Common/SegmentedControl/SegmentedControlView.swift
```diff
@@ -29,19 +29,19 @@ struct SegmentedControlView: View {
                     withAnimation {
                         selectedTabIndex = index
                     }
-                }) {
+                }, label: {
```
justinwongeecs, line 32: For consistency, when possible, we try to use the trailing closure syntax as much as possible. More info on it here: https://www.hackingwithswift.com/sixty/6/5/trailing-closure-syntax. 

So for a button it should look something like this:

```swift
Button(action: {
    ....
}) {
    ....
}
```

justinwongeecs (commented):
I see, I think all the pods changes has to do with the fact with the updated cocoa pods so I think that should be fine?

The changes to the Secrets.swift file is when you added it to your project. I'm wondering if we can do a git ignore of the Pods folder, Secrets.swift, and the GoogleService.plist file.

.gitignore
```diff
@@ -47,8 +47,8 @@ xcuserdata/
 *.moved-aside
 *.xccheckout
 *.xcscmblueprint
-GoogleService-Info.plist
-Secrets.swift
+berkeley-mobile/Secrets.swift
+berkeley-mobile/Secrets.swift
```
justinwongeecs, line 51: Why two `berkeley-mobile/Secrets.swift` lines? Should the google info plist path be here too?

PR #325, opened 2025-02-21
BER-49: Improve SegmentedControlView's Indicator Bar

baeuke:
- Moved the indicator down and changed its height (thickness) so it didn't occlude with the tab names
- Experimented with the numbers for height and spacing.
   - weird behavior happened for indicatorOffset of -30 and indicatorHeight of 8: the tabs were not tappable in the preview (exaclty for these numbers);
   - and for some other numbers the tabs would only be tapped and selected if tapped on the center of the text (tab name) and not the edges
- Small refactoring helped to make the tapping behavior more deterministic
   - reordered the `Text` and `capsule` views in the hierarchy and adjusted the numbers

<div align="center">
Screenshot 2025-02-21 at 12 15 41 PM
</div>


justinwongeecs (approved):
Looks great!

Ohjason-1 (approved):
I like it!

PR #326, opened 2025-02-23
BER-54: Switch to SegmentedControlView

YHC66:
switch the native SwiftUI picker to our custom SegmentedControlView and integrate it with the TabView driving the three separate views in ResourcesView.swift.

### DIFF
--- a/berkeley-mobile/Resources/ResourcesView.swift
+++ b/berkeley-mobile/Resources/ResourcesView.swift
@@ -8,14 +8,16 @@
 
 import SwiftUI
 
-//MARK: ResourcesView
+
+// MARK: ResourcesView
+
 struct ResourcesView: View {
+    @StateObject private var resourcesVM = ResourcesViewModel()
     @State private var tabSelectedValue = 0
     @State private var shoutoutTabSelectedValue = 0
-    @StateObject private var resourcesVM = ResourcesViewModel()
     
     init() {
-        //Use this if NavigationBarTitle is with Large Font
+        // Use this if NavigationBarTitle is with Large Font
         UINavigationBar.appearance().largeTitleTextAttributes = [.font : BMFont.bold(30)]
     }
     
@@ -33,7 +35,7 @@ struct ResourcesView: View {
                        noResourcesAvailableView
                     } else {
                         SegmentedControlView(
-                            tabNames: resourcesVM.resourceCategories.map { $0.name },
+                            tabNames: resourcesVM.resourceCategoryNames,
                             selectedTabIndex: $tabSelectedValue
                         )
                         .padding()
@@ -112,7 +114,9 @@ struct ResourcesView: View {
     }
 }
 
-//MARK: - ResourcePageView
+
+// MARK: - ResourcePageView
+
 struct ResourcePageView: View {
     var resourceSections: [BMResourceSection]
 
@@ -152,7 +156,9 @@ struct ResourcePageView: View {
     }
 }
 
-//MARK: - ResourceLinkView
+
+// MARK: - ResourceLinkView
+
 struct ResourceLinkView: View {
     var resource: BMResource
     
@@ -192,7 +198,9 @@ struct ResourceLinkView: View {
     }
 }
 
-//MARK: - ResourceShoutoutView
+
+// MARK: - ResourceShoutoutView
+
 struct ResourceShoutoutView: View {
     var callout: BMResourceShoutout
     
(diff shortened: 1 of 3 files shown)


justinwongeecs:
Could you add a quick video screen recording of demoing the UI feature to the PR description? If a PR has a UI component, please attach a screenshot or if applicable, a video screen recording.

YHC66:
https://github.com/user-attachments/assets/600a5c89-5ec3-4c93-96b7-6717d382e1d0

attached is the demo

Chhumbucket:
The code looks good. Could u also test clicking on the resources themselves too?

YHC66:
> from

Thanks for your feedback! That also works! Would you want me to provide another video demo showing this?

justinwongeecs:
Screenshot 2025-02-23 at 7 27 00 PM

When I checkouted to your branch, it seems that `GoogleService-Info.plist` is missing for some reason?

berkeley-mobile/Resources/ResourcesView.swift
```diff
@@ -32,12 +32,10 @@ struct ResourcesView: View {
...
                        noResourcesAvailableView
                     } else {
-                        Picker(selection: $tabSelectedValue, label: Text("")) {
-                            ForEach(Array(resourcesVM.resourceCategories.enumerated()), id: \.offset) { idx, category in
-                                Text(category.name).tag(idx)
-                            }
-                        }
-                        .pickerStyle(.segmented)
+                        SegmentedControlView(
+                            tabNames: resourcesVM.resourceCategories.map { $0.name },
```
justinwongeecs, line 36: It might be better to abstract this away into a **computed property** in `ResourcesViewModel` as it is quite verbose at the call site at `SegmentedControlView`. 

Here's an article explaining what computed properties are: https://www.avanderlee.com/swift/computed-property/

berkeley-mobile/Resources/ResourcesView.swift
justinwongeecs, line 1:
```swift
@State private var tabSelectedValue = 0
@State private var shoutoutTabSelectedValue = 0
@StateObject private var resourcesVM = ResourcesViewModel()
```

Put `@StateObject` at the top before all the `@State` variables. By current convention, we put all  `@StateObject`s, `@EnvironmentObjects`, and `@Environment`s at the top before any regular `@State`. When this was written we didn't really have a standard.

justinwongeecs, line 1: Similarly, for comments, put a space before each comment like this `// Comment starts ...`.

For Marks, each Mark should have **two** newlines before it and **one** after it like this:

```swift
}


// MARK: - ResourceShoutoutView

struct ResourceShoutoutView: View { ... }
```

Please update any comments and Marks that are not conforming to this convention.

YHC66:
> Screenshot 2025-02-23 at 7 27 00 PM
> When I checkouted to your branch, it seems that `GoogleService-Info.plist` is missing for some reason?

The `GoogleService-Info.plist` seems to exist in my end? Is it possibly due to some mistakes in git?

berkeley-mobile/Resources/ResourcesView.swift
YHC66: I just made some adjustments to use computed property in ResourcesViewModel.

berkeley-mobile/Resources/ResourcesView.swift
YHC66: fixed!

YHC66: fixed!

berkeley-mobile/Resources/ResourcesViewModel.swift
```diff
@@ -33,11 +33,17 @@ struct BMResourceShoutout: Identifiable, Hashable, Codable {
...
-//MARK: - ResourcesViewModel 
+
+// MARK: - ResourcesViewModel 
+
 class ResourcesViewModel: ObservableObject {
     @Published var shoutouts = [BMResourceShoutout]()
     @Published var resourceCategories = [BMResourceCategory]()
     
+    var resourceCategoryNames: [String] {
+        resourceCategories.map { category in category.name }
```
justinwongeecs, line 44: You can use what you have previously to write it in a more succinct way with shorthand argument labels:

`resourceCategories.map { $0.name }`

YHC66: Got it! Fixed!

justinwongeecs:
Is it possible to remove the changes `project.pbxproj` from this PR? Because the changes there are involving the `GoogleService-Info.plist` and `Secrets.swift` files which have no relevance to this PR?

baeuke (approved):
looks good!

PR #330, opened 2025-02-27
BER-48: Restrict Home Tab Map Viewing Area

baeuke:
- found numbers for the needed region covering ~bay area
   - (latitude: 37.76251429388581, longitude: -122.27164812506234)
   - span: (latitudeDelta: 0.8458437031956976, longitudeDelta: 0.6425468841745499)
   - existing centerMapOnLocation private function (in the same file) uses different values (meters) for span. but since when reading data, ios provides span in latitudeDelta and longitudeDelta, I used those without the approximate conversion.
- set cameraBoundary property of the mapView to that region
- and set a maximum zoom to 300,000 meters (arbitrary - chosen so that there was a little buffer to zoom out from the initial full map region)

https://github.com/user-attachments/assets/a48c1713-b349-4a1d-9220-d8891adc720f


justinwongeecs (approved):
Looks good!

justinwongeecs:
@baeuke looks like there are merge conflicts in `MapViewController.swift`. Checkout to `master` and `git pull` then checkout to `feature/ber-48` and do `git merge master` to fix the merge conflicts.

justinwongeecs (approved):
Looks great!

PR #340, opened 2025-03-10
BER-72: Rewrite CardView in SwiftUI as a View Modifier

iychi:
ticket: https://linear.app/moon-stone/issue/BER-72/rewrite-cardview-in-swiftui-as-a-view-modifier
I've written the view modifier for Cardify in View+Extension.swift and test it.

cardify test

### DIFF
--- a/berkeley-mobile/Utils/View+Extension.swift
+++ b/berkeley-mobile/Utils/View+Extension.swift
@@ -30,7 +30,7 @@ struct HomeMapControlButtonStyle: ButtonStyle {
 struct Cardify: ViewModifier {
     func body(content: Content) -> some View {
         content
-            .padding()
+            .padding(3)
             .background(
                 RoundedRectangle(cornerRadius: 12)
                     .fill(Color(BMColor.cardBackground)) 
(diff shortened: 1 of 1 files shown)


Chhumbucket (approved):
Code looks good

berkeley-mobile/Utils/View+Extension.swift
```diff
@@ -25,3 +25,22 @@ struct HomeMapControlButtonStyle: ButtonStyle {
             )
     }
 }
+
+
```
baeuke, line 29: I think, since we don't have a stored property, you can just directly extend View without having a separate struct for the modifier. that way can be less redundant.

baeuke (commented):
little refactor?

berkeley-mobile/Utils/View+Extension.swift
justinwongeecs: The current syntax is preferred as it's used by Apple: [https://developer.apple.com/documentation/swiftui/viewmodifier](url)

It's indeed a bit more verbose, but there's a clear separation of concern between the extension on the View and the struct conforming to the `ViewModifier` protocol.

baeuke (approved):
good to go!

PR #343, opened 2025-03-17
BER-73: Rewrite OpenTimesCardView in SwiftUI

YHC66:
- Rewrote OpenTimesCardView in SwiftUI
- Added expandable/collapsible functionality with chevron animation
- Displays open/closed status with color indicators
- Shows day-by-day hours when expanded

Currently test it in preview in OpenTimesCardSwiftUI

image

https://github.com/user-attachments/assets/23fb46dd-199c-4470-bd32-6a7419298efa


berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
```diff
@@ -0,0 +1,251 @@
...
+            size = newSize
+            onSizeChange?(newSize)
+        }
+    }
+}
+
+struct OpenTimesCard: View {
+    let item: HasOpenTimes
+    @State private var isExpanded = false
+    @EnvironmentObject private var sizeReporter: SizeReporter
```
justinwongeecs, line 25: Order `@EnvironmentObject`s first then any `@State`s then any `@Binding`s then regular parameters/properties. with a newline separating each "category", though you can group any `@State` and `@Binding` variables together. This way it's easier to distinguish what is what.

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
```diff
@@ -0,0 +1,251 @@
...
+//      
+//  Created by Yihang Chen on 3/13/25.
+//  Copyright © 2025 ASUC OCTO. All rights reserved.
+//
+
+import SwiftUI
+
+class SizeReporter: ObservableObject {
+    @Published var size: CGSize = .zero
+    var onSizeChange: ((CGSize) -> Void)?
```
justinwongeecs, line 12: Add a newline separating any `@Published` variables and regular properties.

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
```diff
@@ -0,0 +1,251 @@
...
+
+struct SizePreferenceKey: PreferenceKey {
+    static var defaultValue: CGSize = .zero
+    
+    static func reduce(value: inout CGSize, nextValue: () -> CGSize) {
+        value = nextValue()
+    }
+}
+
+// MARK: - Preview
```
justinwongeecs, line 217: Add an additional newline. There should be two newlines before a MARK.

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
```diff
@@ -0,0 +1,251 @@
...
+    private func toggleWithAnimation() {
+        withAnimation(.spring(response: 0.4, dampingFraction: 1.0)) {
+            isExpanded.toggle()
+        }
+    }
+    
+    private func hoursView(for day: DayOfWeek, isCurrentDay: Bool) -> some View {
+        let defaultText = Text("Closed")
+            .font(.system(size: 12, weight: isCurrentDay ? .bold : .regular))
+            .foregroundColor(Color(.darkGray))
```
justinwongeecs, line 161: We should not be returning a SwiftUI view from a function here. If you notice, this looks very similar to what you are returning in this function on lines 186-188. The `Text` view should be located in the view's body, and you should call these helper functions as arguments into `Text` so this function should be returning some String instead. 

This also applies to `formatTimeInterval(_:)`. Return a `String` instead of a `Text`. The return value (the SwiftUI `Text` is also very similar to what we are returning in `hoursView(for:isCurrentDay:)`

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
```diff
@@ -0,0 +1,251 @@
...
+            )
+            .onPreferenceChange(SizePreferenceKey.self) { newSize in
+                if isExpanded {
+                    withAnimation(.spring(response: 0.6, dampingFraction: 1.0)) {
+                        sizeReporter.updateSize(newSize)
+                    }
+                } else {
+                    sizeReporter.updateSize(newSize)
+                }
+            }
```
justinwongeecs, line 65: What's the purpose of `SizeReporter` and the usage of Preferences here. Preferences are used to pass data up from a subview to its parent view. I'm not quite sure why the `SizeReporter` is necessary? And what's this `Color.clear` overlay used for?

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
```diff
@@ -0,0 +1,251 @@
...
+        }
+    }
+    
+    @ViewBuilder
+    private func headerView() -> some View {
+        Button(action: {
+            toggleWithAnimation()
+        }) {
+            HStack(spacing: 12) { 
+                Image("Clock")
```
justinwongeecs, line 75: Don't use the "Clock" image in the Assets folder. Swap out for an SF Symbol one as that comes with automatic light/dark mode support. After you swapped it out, double check if we are using the "Clock" local asset anywhere else. If we aren't remove it from Assets.

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
```diff
@@ -0,0 +1,251 @@
...
+                
+                HStack {
+                    if let isOpen = item.isOpen {
+                        Text(isOpen ? "Open" : "Closed")
+                            .font(.system(size: 12, weight: .medium)) 
+                            .padding(.horizontal, 12) 
+                            .padding(.vertical, 4) 
+                            .background(isOpen ? Color.green : Color.blue.opacity(0.7)) 
+                            .foregroundColor(.white)
+                            .cornerRadius(12)
```
justinwongeecs, line 89: Is it possible to align the "Open"/"Closed" text with the rest of the "Open"/"Closed" text in the dropdown body when expanded. Right now, they are not aligned and visually I think it looks a bit off?

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
```diff
@@ -0,0 +1,251 @@
...
+        }
+    }
+    
+    @ViewBuilder
+    private func expandableContentView() -> some View {
+        VStack(spacing: 2) { 
+            ForEach(0..<dayNames.count, id: \.self) { index in
+                let day = dayNames[index]
+                let dayOfWeek = DayOfWeek.allCases[index]
+                let isCurrentDay = index == currentDayOfWeek
```
justinwongeecs, line 127: Why is `currentDayOfWeek` a computed property if we are only using it once here?

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
```diff
@@ -0,0 +1,251 @@
...
+                
+                HStack() {
+                    Text(day)
+                        .font(.system(size: 12, weight: isCurrentDay ? .bold : .regular)) 
+                        .foregroundColor(.black)
+                        .padding(.leading, 55)
+                        .lineLimit(1)
```
justinwongeecs, line 134: Move `.lineLimit(1)`to the `HStack` since you're using it on both subviews

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
```diff
@@ -0,0 +1,251 @@
...
+    
+    @ViewBuilder
+    private func expandableContentView() -> some View {
+        VStack(spacing: 2) { 
+            ForEach(0..<dayNames.count, id: \.self) { index in
+                let day = dayNames[index]
+                let dayOfWeek = DayOfWeek.allCases[index]
+                let isCurrentDay = index == currentDayOfWeek
+                
+                HStack() {
```
justinwongeecs, line 129: Delete extraneous `()` on `HStack()`

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
```diff
@@ -0,0 +1,251 @@
...
+                        .lineLimit(1) 
+                        .lineSpacing(0) 
+                        .frame(maxWidth: .infinity, alignment: .trailing)
+                }
+                .frame(height: 22) 
+                .contentShape(Rectangle())
+            }
+        }
+        .padding(.vertical, 6)
+        .background(Color.white)
```
justinwongeecs, line 149: To keep it consistent with the rest of the file, change `Color.white` to `.white`. Swift can infer it.

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
```diff
@@ -0,0 +1,251 @@
...
+        func nextOpenInterval() -> HoursInterval? {
+            return nil
+        }
+        
+        static func createEmptyWeeklyHours() -> WeeklyHours {
+            let weeklyHours = WeeklyHours()
+            return weeklyHours
+        }
+    }
+}
```
justinwongeecs, line 250: In Previews, could you also show an instance where it displays sample data so it's not always closed to test what it looks like if the weekly hours are not empty?

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
YHC66: That makes sense! Turned formatTimeInterval and hourView to return String now

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
YHC66: makes sense!

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
YHC66: Sounds good! Will do

YHC66: Do you mean commenting out the previous preview and adding a new preview that shows the weekly hours?

justinwongeecs: You can display another view in the `PreviewProvider` to show two Simulators. Check it out here: https://developer.apple.com/documentation/swiftui/previewprovider

YHC66: Oh! Got it!

YHC66: image
I'm not sure if this looks good.

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
YHC66: Using Color.clear allows measurement without affecting appearance. This along with the SizeReporter creates a notification system that alerts parent views when the card's dimensions change, allowing them to adjust their layouts appropriately while maintaining smooth animations.

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
YHC66: Changed! However, the SF Symbol clock is not the same as what it shows in Figma. But it looks still fine.

YHC66: Do you mean I should remove the clock local assets for all the files using it and replace them as the SF symbol clock?

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
YHC66: image
image
Fixed!

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
justinwongeecs: Yup! Also, you don't need to follow the Figma exactly. Feel free to play around with it. For using symbols from now on, we will be using SF Symbols

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
justinwongeecs: On a second look, what is the time/text to the left of the chevron in the dropdown header represent? Like for the right screenshot in the dropdown header, you have two "Closed" which is kind of ambiguous

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
justinwongeecs: Is it possible to achieve the same without using this mechanism?

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
justinwongeecs: Yup that looks good!

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
```diff
@@ -0,0 +1,295 @@
...
+    private func headerView() -> some View {
+        Button(action: {
+            toggleWithAnimation()
+        }) {
+            HStack(spacing: 12) { 
+                Image(systemName: "clock.fill") 
+                    .resizable()
+                    .aspectRatio(contentMode: .fit)
+                    .frame(width: 18, height: 18) 
+                    .foregroundColor(Color.gray)
```
justinwongeecs, line 77: `Color.gray` &rarr; `.gray`

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
```diff
@@ -0,0 +1,295 @@
...
+                
+                HStack {
+                    if let isOpen = item.isOpen {
+                        Text(isOpen ? "Open" : "Closed")
+                            .font(.system(size: 12, weight: .medium)) 
+                            .padding(.horizontal, 12) 
+                            .padding(.vertical, 4) 
+                            .background(isOpen ? Color.blue : Color.blue.opacity(0.7))
```
justinwongeecs, line 85: `Color.blue` &rarr;  `.blue`

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
```diff
@@ -0,0 +1,295 @@
...
+                        Text(formatTimeIntervalText(nextOpenInterval.dateInterval))
+                            .foregroundColor(Color(.darkGray))
+                            .font(.system(size: 12, weight: nextOpenInterval.dateInterval.contains(Date()) ? .bold : .regular))
+                            .padding(.trailing, 0)
+                    } else {
+                        Text("Closed")
+                            .font(.system(size: 12, weight: .regular)) 
+                            .foregroundColor(Color(.darkGray))
+                            .padding(.trailing, 0)
+                    }
```
justinwongeecs, line 103: A lot of shared code. You can rewrite it with just one `Text` view and use ternary operators to appropriate display the attributes and values based on if `item.nextOpenInterval()` is `nil` or not.

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
```diff
@@ -0,0 +1,295 @@
...
+                }
+                
+                Image(systemName: "chevron.down")
+                    .foregroundColor(Color.gray)
```
justinwongeecs, line 107: `Color.gray` &rarr; `.gray`

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
```diff
@@ -0,0 +1,295 @@
...
+            .padding(.vertical, 10) 
+        }
+        .buttonStyle(PlainButtonStyle())
+        
+        if isExpanded {
+            Divider()
+        }
+    }
+    
+    @ViewBuilder
```
justinwongeecs, line 124: Remove `@ViewBuilder`. You don't need it now

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
```diff
@@ -0,0 +1,295 @@
...
+                        .lineSpacing(0)
+                        .padding(.trailing, 26)
+                }
+                .frame(height: 22) 
+                .contentShape(Rectangle())
+                .lineLimit(1)
+            }
+        }
+        .padding(.vertical, 6)
+        .background(.white)
```
justinwongeecs, line 154: Did the original use a `BMColor` for the background?

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
```diff
@@ -0,0 +1,295 @@
...
+        }
+        
+        let formatter = DateIntervalFormatter()
+        formatter.dateStyle = .none
+        formatter.timeStyle = .short
+        
+        guard let start = interval.dateInterval.start.timeOnly(),
+              let end = interval.dateInterval.end.timeOnly() else {
+            return "Closed"
+        }
```
justinwongeecs, line 184: You can combine the two `guard` statements together

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
```diff
@@ -0,0 +1,295 @@
...
+        let formatter = DateIntervalFormatter()
+        formatter.dateStyle = .none
+        formatter.timeStyle = .short
+        
+        if let start = interval.start.timeOnly(),
+           let end = interval.end.timeOnly() {
+            return formatter.string(from: start, to: end)
+        } else {
+            return ""
+        }
```
justinwongeecs, line 199: Rewrite into `guard` statement for more cleaner code?

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
```diff
@@ -0,0 +1,295 @@
...
+                .previewDisplayName("Closed Item")
+            
+            // Preview 2: Open
+            OpenTimesCard(item: OpenItem())
+                .environmentObject(SizeReporter())
+                .frame(width: 350)
+                .background(Color(UIColor.systemGroupedBackground))
+                .padding()
+                .previewDisplayName("Open Item")
+        }
```
justinwongeecs, line 232: For any shared view modifiers, put them on the `Group`.

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
```diff
@@ -0,0 +1,295 @@
...
+    // Close hour sample item 
+    struct OpenItem: HasOpenTimes {
+        var weeklyHours: WeeklyHours? = createSampleWeeklyHours()
+        
+        var isOpen: Bool? {
+            return true
+        }
+        
+        func nextOpenInterval() -> HoursInterval? {
+
```
justinwongeecs, line 259: Remove extraneous newline

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
```diff
@@ -0,0 +1,295 @@
...
+        }
+    }
+    
+    // Close hour sample item 
+    struct OpenItem: HasOpenTimes {
+        var weeklyHours: WeeklyHours? = createSampleWeeklyHours()
+        
+        var isOpen: Bool? {
+            return true
+        }
```
justinwongeecs, line 256: Why a computed property that always return the exact same value?

Chhumbucket (changes requested):
Fix with Justin edits

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
YHC66: I tried to rely on SwiftUI's natural layout system rather than manually tracking sizes using `SizeReporter`, which seems to solve the problem.

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
YHC66: Was refering to figma design, but yeah, will delete that!

YHC66: image

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
YHC66: Turned .white to BMColor.cardBackground

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
YHC66: Combined!

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
YHC66: Fixed

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
YHC66: fixed!

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
justinwongeecs: If it doesn't work then sure. 
Sorry, what does "tend to appear in the same preview page" mean?

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
```diff
@@ -0,0 +1,260 @@
...
+//  Created by Yihang Chen on 3/13/25.
+//  Copyright © 2025 ASUC OCTO. All rights reserved.
+//
+
+import SwiftUI
+
+struct OpenTimesCard: View {
+    let item: HasOpenTimes
+    
+    @State private var isExpanded = false
```
justinwongeecs, line 13: Usually we have our `@State` variables before our regular variables.

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
```diff
@@ -0,0 +1,260 @@
...
+    
+    private let dayNames = ["Sunday", "Monday", "Tuesday", "Wednesday", "Thursday", "Friday", "Saturday"]
+    
+    var body: some View {
+        VStack(spacing: 0) {
```
justinwongeecs, line 18: `VStack(spacing: 0) { ... }` &rarr; `VStack { ... }`

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
```diff
@@ -0,0 +1,260 @@
...
+                            .frame(width: 80, alignment: .leading)
+                    }
+                    
+                    Spacer()
+                    
+                    if let nextOpenInterval = item.nextOpenInterval() {
+                        Text(formatTimeIntervalText(nextOpenInterval.dateInterval))
+                            .foregroundColor(Color(.darkGray))
+                            .font(.system(size: 12, weight: nextOpenInterval.dateInterval.contains(Date()) ? .bold : .regular))
+                            .padding(.trailing, 0)
```
justinwongeecs, line 66: Is this not necessary?

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
```diff
@@ -0,0 +1,260 @@
...
+                        
+                    
+                    Spacer()
+                    
+                    Text(hoursText(for: dayOfWeek, isCurrentDay: isCurrentDay))
+                        .font(.system(size: 12, weight: isCurrentDay ? .bold : .regular))
+                        .foregroundColor(Color(.darkGray))
+                        .lineSpacing(0)
+                        .padding(.trailing, 26)
+                }
```
justinwongeecs, line 111: 1. move `.lineSpacing(0)` to `HStack`
2. move .font(...)` to `HStack`

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
```diff
@@ -0,0 +1,260 @@
...
+            }
+        }
+    }
+}
+
+extension View {
+    func positionedAtTop() -> some View {
+        self.modifier(TimeCardTopModifier())
+    }
+}
```
justinwongeecs, line 186: Move all custom view modifiers code to `View+Extension.swift` file.

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
```diff
@@ -0,0 +1,260 @@
...
+                .positionedAtTop()
+                .previewLayout(.sizeThatFits)
+                .previewDisplayName("Closed Item")
+        
+        OpenTimesCard(item: OpenItem())
+            .positionedAtTop()
+            .previewLayout(.sizeThatFits)
+            .previewDisplayName("Open Item")
+        }
+    }
```
justinwongeecs, line 204: Could you rewrite `struct OpenTimesCard_Previews: PreviewProvider` using the `#Preview` macro?

If you want to have two separate simulators for the closed item and one for the open item you will need to create two separate `#Preview`s like this: 


```swift
#Preview("Closed Item Card") {
  ...
}
 
 #Preview("Open Item Card") {
  ...
 }
```

berkeley-mobile/Fitness/GymDetailView.swift
justinwongeecs, line 1: Delete comment

justinwongeecs (changes requested):

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
YHC66: will remember that!

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
```diff
@@ -0,0 +1,227 @@
...
+
+
+// MARK: - Previews
+
+#Preview("Closed Item Card") {
+    OpenTimesCard(item: ClosedItem())
+        .positionedAtTop()
+        .previewLayout(.sizeThatFits)
```
justinwongeecs, line 218: Remove. We use `.previewLayout(...)` when we write Previews using `PreviewProvider`. With `#Preview` macro, we can achieve the same effect with using the `traits` parameter like this: 

```swift
#Preview(traits: .sizeThatFitsLayout) {
   ...
}
```

However, this is only available in iOS 17.0+ so we can't use it just yet.

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
```diff
@@ -0,0 +1,227 @@
...
+}
+ 
+#Preview("Open Item Card") {
+    OpenTimesCard(item: OpenItem())
+        .positionedAtTop()
+        .previewLayout(.sizeThatFits)
```
justinwongeecs, line 224: Remove.

berkeley-mobile/Fitness/GymDetailView.swift
justinwongeecs, line 1: Delete file from this commit.

berkeley-mobile/Utils/View+Extension.swift
```diff
@@ -27,6 +27,25 @@ struct BMControlButtonStyle: ButtonStyle {
 }
 
 
+// MARK: - OpenTimesCard Positioning
+
+struct OpenTimesCardTopModifier: ViewModifier {
+    func body(content: Content) -> some View {
+        ZStack(alignment: .top) {
+            Color(.systemGroupedBackground)
+                .edgesIgnoringSafeArea(.all)
```
justinwongeecs, line 36: `.edgesIgnoringSafeArea(...)` is depreciated. Use `.ignoresSafeArea(...)` instead

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
```diff
@@ -0,0 +1,227 @@
...
+                        Text(formatTimeIntervalText(nextOpenInterval.dateInterval))
+                            .foregroundColor(Color(.darkGray))
+                            .font(.system(size: 12, weight: nextOpenInterval.dateInterval.contains(Date()) ? .bold : .regular))
+                    } 
+                }
+                
+                Image(systemName: "chevron.down")
+                    .foregroundColor(.gray)
+                    .font(.system(size: 12, weight: .medium)) 
+                    .rotationEffect(.degrees(isExpanded ? 180 : 0))
```
justinwongeecs, line 72: When it is not expanded, the chevron should be facing to the right like this 
chevron forward
. When it is expanded it should animate by turning clockwise to result it pointing downwards like this 
chevron down.

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
```diff
@@ -0,0 +1,227 @@
...
+        .background(Color(BMColor.cardBackground))
+        .cornerRadius(15)
+        .shadow(color: .black.opacity(0.05), radius: 3, x: 0, y: 1)
+        .animation(.spring(response: 0.4, dampingFraction: 1.0), value: isExpanded)
+        .padding(.bottom, 8) 
+        .zIndex(100)
+    }
+    
+    @ViewBuilder
+    private func headerView() -> some View {
```
justinwongeecs, line 37: Refactor smaller subviews in `headerView` into private variables

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
```diff
@@ -0,0 +1,227 @@
...
+            return ""
+        }
+        
+        return formatter.string(from: start, to: end)
+    }
+}
+
+    struct ClosedItem: HasOpenTimes {
+        var weeklyHours: WeeklyHours? = createEmptyWeeklyHours()
+        var isOpen: Bool? = false
```
justinwongeecs, line 161: Shouldn't we be using the `isOpen` from `HasOpenTimes`. We don't need to create a new separate one.

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
```diff
@@ -0,0 +1,227 @@
...
+        
+        static func createEmptyWeeklyHours() -> WeeklyHours {
+            let weeklyHours = WeeklyHours()
+            return weeklyHours
+        }
+    }
+    
+    struct OpenItem: HasOpenTimes {
+        var weeklyHours: WeeklyHours? = createSampleWeeklyHours()
+        var isOpen: Bool? = true
```
justinwongeecs, line 175: Like above, we should be using `isOpen` already in `HasOpenTimes`. So we can remove this property.

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
```diff
@@ -0,0 +1,227 @@
...
+                    .frame(width: 18, height: 18) 
+                    .foregroundColor(.gray)
+                
+                HStack {
+                    if let isOpen = item.isOpen {
+                        Text(isOpen ? "Open" : "Closed")
+                            .font(.system(size: 12, weight: .medium)) 
+                            .padding(.horizontal, 12) 
+                            .padding(.vertical, 4) 
+                            .background(isOpen ? .blue : .blue.opacity(0.7))
```
justinwongeecs, line 54: I think changing the background color depending if it is closed or open would be visually more informative. For example, maybe a red color when it is closed and green if open? 

Any thoughts?

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
```diff
@@ -0,0 +1,227 @@
...
+            
+            func createHoursInterval(day: DayOfWeek, startHour: Int, startMinute: Int = 0, endHour: Int, endMinute: Int = 0) {
+                let today = Date()
+                let calendar = Calendar.current
+                
+                let todayComponents = calendar.dateComponents([.year, .month, .day], from: today)
+                guard let baseDate = calendar.date(from: todayComponents) else { return }
+                
+                guard let startDate = calendar.date(bySettingHour: startHour, minute: startMinute, second: 0, of: baseDate),
+                      let endDate = calendar.date(bySettingHour: endHour, minute: endMinute, second: 0, of: baseDate) else { return }
```
justinwongeecs, line 195: I think you can check out `getTodayShiftDate(for:hourComponent:minuteComponent:secondComponent)` in `Date+Extension`?

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
```diff
@@ -0,0 +1,227 @@
...
+                weeklyHours.addInterval(interval, to: day)
+            }
+            
+            // Sample hours
+            createHoursInterval(day: .monday, startHour: 9, endHour: 17)      // 9 AM - 5 PM
+            createHoursInterval(day: .tuesday, startHour: 9, endHour: 17)     // 9 AM - 5 PM
+            createHoursInterval(day: .wednesday, startHour: 9, endHour: 17)   // 9 AM - 5 PM
+            createHoursInterval(day: .thursday, startHour: 9, endHour: 17)    // 9 AM - 5 PM
+            createHoursInterval(day: .friday, startHour: 9, endHour: 16)      // 9 AM - 4 PM
+            createHoursInterval(day: .saturday, startHour: 10, endHour: 14)   // 10 AM - 2 PM
```
justinwongeecs, line 207: For the open time, make all the `startHour` and `endHour` the maximum interval so that we can always see "Open"

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
```diff
@@ -0,0 +1,227 @@
...
+        
+        static func createSampleWeeklyHours() -> WeeklyHours {
+            let weeklyHours = WeeklyHours()
+            
+            func createHoursInterval(day: DayOfWeek, startHour: Int, startMinute: Int = 0, endHour: Int, endMinute: Int = 0) {
+                let today = Date()
+                let calendar = Calendar.current
+                
+                let todayComponents = calendar.dateComponents([.year, .month, .day], from: today)
+                guard let baseDate = calendar.date(from: todayComponents) else { return }
```
justinwongeecs, line 192: Per the Formatting Guide, the `return` keyword should be a newline like this: 

```swift
guard [CONDITION] else {
    return
}
```

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
```diff
@@ -0,0 +1,227 @@
...
+        .zIndex(100)
+    }
+    
+    @ViewBuilder
+    private func headerView() -> some View {
+        Button(action: {
+            toggleWithAnimation()
+        }) {
+            HStack(spacing: 12) { 
+                Image(systemName: "clock.fill")
```
justinwongeecs, line 42: Use the "clock" symbol, not the fill variant

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
```diff
@@ -0,0 +1,227 @@
...
+    private func headerView() -> some View {
+        Button(action: {
+            toggleWithAnimation()
+        }) {
+            HStack(spacing: 12) { 
+                Image(systemName: "clock.fill") 
+                    .resizable()
+                    .aspectRatio(contentMode: .fit)
+                    .frame(width: 18, height: 18) 
+                    .foregroundColor(.gray)
```
justinwongeecs, line 46: Remove. We will be using the default `.primary` foregroundColor

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
```diff
@@ -0,0 +1,227 @@
...
+                    
+                    if let nextOpenInterval = item.nextOpenInterval() {
+                        Text(formatTimeIntervalText(nextOpenInterval.dateInterval))
+                            .foregroundColor(Color(.darkGray))
+                            .font(.system(size: 12, weight: nextOpenInterval.dateInterval.contains(Date()) ? .bold : .regular))
+                    } 
+                }
+                
+                Image(systemName: "chevron.down")
+                    .foregroundColor(.gray)
```
justinwongeecs, line 70: Remove.

justinwongeecs (changes requested):

berkeley-mobile/Fitness/GymDetailView.swift
YHC66: Will delete it! I also need to modify project.pbxproj file right?

justinwongeecs: Yup! It should automatically change the `project.pbxproj` file. The `project.pbxproj` keeps track of changes to the project like adding/removing files, renaming of files, etc...

justinwongeecs (changes requested):
There's still some comments that need to be resolved previously.

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
YHC66: Yep! I think red when closed and green when open works!

YHC66:
> There's still some comments that need to be resolved previously.

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
```diff
@@ -0,0 +1,240 @@
...
+    
+    private var clockIcon: some View {
+        Image(systemName: "clock") 
+            .resizable()
+            .aspectRatio(contentMode: .fit)
+            .frame(width: 18, height: 18) 
+    }
+    
+    private var statusBadge: some View {
+        Group {
```
justinwongeecs, line 73: Remove `Group` and use `@ViewBuilder` instead. This avoids having to add an additional nesting/indentation to the code.

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
```diff
@@ -0,0 +1,240 @@
...
+                    .background(isOpen ? Color.green : Color.red) 
+                    .foregroundColor(.white)
+                    .cornerRadius(12)
+                    .frame(width: 80, alignment: .leading)
+            }
+        }
+    }
+    
+    private var openTimeText: some View {
+        Group {
```
justinwongeecs, line 88: Remove `Group`. Use `@ViewBuilder` instead.

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
```diff
@@ -0,0 +1,240 @@
...
+            if let nextOpenInterval = item.nextOpenInterval() {
+                Text(formatTimeIntervalText(nextOpenInterval.dateInterval))
+                    .foregroundColor(Color(.darkGray))
```
justinwongeecs, line 91: Remove foregroundColor. In dark mode, the color does not contrast well: 
Screenshot 2025-03-23 at 10 20 18 PM

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
```diff
@@ -0,0 +1,240 @@
...
+    private func expandableContentView() -> some View {
+        VStack(spacing: 2) { 
+            ForEach(0..<dayNames.count, id: \.self) { index in
+                let day = dayNames[index]
+                let dayOfWeek = DayOfWeek.allCases[index]
+                let isCurrentDay = index == Date().weekday()
+                
+                HStack {
+                    Text(day)
+                        .foregroundColor(.black)
```
justinwongeecs, line 115: Remove `foregroundColor`. Color doesn't contrast well in dark mode: 
Screenshot 2025-03-23 at 10 21 48 PM

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
```diff
@@ -0,0 +1,240 @@
...
+                        .padding(.leading, 55)
+                    
+                    Spacer()
+                    
+                    Text(hoursText(for: dayOfWeek, isCurrentDay: isCurrentDay))
+                        .foregroundColor(Color(.darkGray))
```
justinwongeecs, line 121: Remove `foregroundColor`.

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
```diff
@@ -0,0 +1,240 @@
...
+                    .font(.system(size: 12, weight: nextOpenInterval.dateInterval.contains(Date()) ? .bold : .regular))
+            }
+        }
+    }
+    
+    private var chevronIcon: some View {
+        Image(systemName: "chevron.right")
+            .font(.system(size: 12, weight: .medium)) 
+            .rotationEffect(.degrees(isExpanded ? 90 : 0))
+            .animation(isExpanded ? nil : .spring(response: 0.6, dampingFraction: 1.0), value: isExpanded)
```
justinwongeecs, line 101: If you notice, when `isExpanded` changes from `false` to `true`, the chevron doesn't animate since it's set to `nil`: 

https://github.com/user-attachments/assets/9ac3bda7-eb77-49f5-84ce-88217d157916

Remove the ternary operator and just have the spring animation.

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
```diff
@@ -0,0 +1,240 @@
...
+            }
+            
+            // Sample hours
+            createHoursInterval(day: .monday, startHour: 0, startMinute: 0, endHour: 23, endMinute: 59)
+            createHoursInterval(day: .tuesday, startHour: 0, startMinute: 0, endHour: 23, endMinute: 59)
+            createHoursInterval(day: .wednesday, startHour: 0, startMinute: 0, endHour: 23, endMinute: 59)
+            createHoursInterval(day: .thursday, startHour: 0, startMinute: 0, endHour: 23, endMinute: 59)
+            createHoursInterval(day: .friday, startHour: 0, startMinute: 0, endHour: 23, endMinute: 59)
+            createHoursInterval(day: .saturday, startHour: 0, startMinute: 0, endHour: 23, endMinute: 59)
+            createHoursInterval(day: .sunday, startHour: 0, startMinute: 0, endHour: 23, endMinute: 59)
```
justinwongeecs, line 222: Use a for loop instead?

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
```diff
@@ -0,0 +1,240 @@
...
+    OpenTimesCard(item: ClosedItem())
+        .positionedAtTop()
+}
+ 
+#Preview("Open Item Card") {
+    OpenTimesCard(item: OpenItem())
+        .positionedAtTop()
+}
+
+
```
justinwongeecs, line 240: Remove unnecessary newlines.

justinwongeecs (changes requested):
Make sure to also check your UI implementation in Dark Mode to ensure text legibility and contrast. In the SwiftUI Canvas, you can click on the "Canvas Device Settings" menu to manually select the Color Scheme:
Screenshot 2025-03-23 at 10 28 41 PM

YHC66:
image
image
Thanks for spotting the issues!

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
```diff
@@ -0,0 +1,229 @@
...
+//
+
+import SwiftUI
+
+struct OpenTimesCard: View {
+    @State private var isExpanded = false
+
+    let item: HasOpenTimes
+    
+    private let dayNames = ["Sunday", "Monday", "Tuesday", "Wednesday", "Thursday", "Friday", "Saturday"]
```
justinwongeecs, line 15: Perhaps this is better situated in `DayOfWeek` instead? 
It would be nice to have each case to have its respective string like `sunday` to `"Sunday"`, `monday` to `"Monday"`, etc...

Check out `CustomStringConvertible`: https://developer.apple.com/documentation/Swift/CustomStringConvertible
to make a type have its own custom textual representation. 

Then perhaps you can have a property where you can get a list of all of the day of week strings.

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
```diff
@@ -0,0 +1,229 @@
...
+        formatter.dateStyle = .none
+        formatter.timeStyle = .short
+        
+        guard let start = interval.start.timeOnly(),
+              let end = interval.end.timeOnly() else {
+            return ""
+        }
+        
+        return formatter.string(from: start, to: end)
+    }
```
justinwongeecs, line 169: The code highlighted here shares very similar logic. I think you can refactor this into a separate function that takes in a String and returns it whenever the guard statement fails.

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
```diff
@@ -0,0 +1,229 @@
...
+            let weeklyHours = WeeklyHours()
+            return weeklyHours
+        }
+    }
+    
+    struct OpenItem: HasOpenTimes {
+        var weeklyHours: WeeklyHours? = createSampleWeeklyHours()
+        
+        func nextOpenInterval() -> HoursInterval? {
+            guard let weeklyHours = weeklyHours else {
```
justinwongeecs, line 189: `guard let weeklyHours = weeklyHours` &rarr; `guard let weeklyHours`

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
```diff
@@ -0,0 +1,229 @@
...
+    struct ClosedItem: HasOpenTimes {
+        var weeklyHours: WeeklyHours? = createEmptyWeeklyHours()
+        
+        func nextOpenInterval() -> HoursInterval? {
+            return nil
+        }
+        
+        static func createEmptyWeeklyHours() -> WeeklyHours {
+            let weeklyHours = WeeklyHours()
+            return weeklyHours
```
justinwongeecs, line 181: why not just `return WeeklyHours()`?

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUI.swift
```diff
@@ -0,0 +1,229 @@
+//
+//  OpenTimesCardSwiftUI.swift
+//      
+//  Created by Yihang Chen on 3/13/25.
+//  Copyright © 2025 ASUC OCTO. All rights reserved.
+//
+
+import SwiftUI
+
+struct OpenTimesCard: View {
```
justinwongeecs, line 10: Usually for SwiftUI views, our naming convention is `...View`. So it should be something like `OpenTimesCardSwiftUIView` for now. 

Please also match the file name to this too.

berkeley-mobile/Utils/View+Extension.swift
```diff
@@ -40,6 +59,10 @@ struct Cardify: ViewModifier {
 }
 
 extension View {
+    func positionedAtTop() -> some View {
+        self.modifier(OpenTimesCardTopModifier())
+    }
```
justinwongeecs, line 64: This is a bit confusing. It seems like `positionedAtTop()` can be applied to any view right? But the underlying ViewModifier is titled as `OpenTimesCardTopModifier` which is very specific. 

Perhaps a more appropriate name than `OpenTimesCardTopModifier` would be something like `PositionAtTopModifier`?

justinwongeecs (changes requested):

berkeley-mobile/Utils/View+Extension.swift
YHC66: Sounds good!

berkeley-mobile/Utils/DayOfWeek.swift
```diff
@@ -18,6 +18,14 @@ enum DayOfWeek: Int, CaseIterable {
     case friday    = 5
     case saturday  = 6
     
+    var description: String {
+        return stringRepresentation()
+    }
+
```
justinwongeecs, line 24: This does indeed work! But I think it would be more suitable to have the `description` not depend on `stringRepresentation`. It's kinda of weird deriving `description` soley based on some other function. In other words, the logic of switching upon `self` and returning the corresponding String should be in `description`. So, 
```swift
switch self {
        case .sunday:
            return "Sunday"
        case .monday:
            return "Monday"
        case .tuesday:
            return "Tuesday"
        case .wednesday:
            return "Wednesday"
        case .thursday:
            return "Thursday"
        case .friday:
            return "Friday"
        case .saturday:
            return "Saturday"
        }
```

should be in `description` and we should use `description` within `stringRepresentation(includesYesterdayTodayVerbage:)`.

justinwongeecs (changes requested):

justinwongeecs (changes requested):

berkeley-mobile/Utils/DayOfWeek.swift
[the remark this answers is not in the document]
YHC66: Sounds good! It turns out I can just return String(description.prefix(1)) for charRepresentation()

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUIView.swift
```diff
@@ -0,0 +1,224 @@
...
+    
+    private let timeFormatter: DateIntervalFormatter = {
+        let formatter = DateIntervalFormatter()
+        formatter.dateStyle = .none
+        formatter.timeStyle = .short
+        return formatter
+    }()
+    
+    var body: some View {
+        VStack {
```
justinwongeecs, line 23: One more thing. Change to `VStack(spacing: 0)`. Sorry you might had this before. But what I realized was that SwiftUI was adding default padding between the divider and the rest of the header view so it resulted in some extra padding above the Divider. To fix this so that the padding is uniform, we need to specify spacing to be zero. 

In the future, we should always ask SwiftUI to infer the padding with just `VStack { ... }` but in cases like this we have to specifically tell the `VStack` to remove the default padding. Sorry about the confusion. 


https://github.com/user-attachments/assets/145f42e6-f7ac-485b-8feb-857f750a956b

justinwongeecs (changes requested):

berkeley-mobile/Common/DetailView/OpenTimesCardSwiftUIView.swift
YHC66: Thanks for clarification!

justinwongeecs (approved):
😎

Chhumbucket (approved):

PR #345, opened 2025-03-18
BER-66: Update Home Map User Location Button Based On Location Permissions

justinwongeecs:
- If user has disabled location permissions, we switch to the `location.slash` SF Symbol variant 
- Rename `LocationManager` to `BMLocationManager`
- Update depreciated `locationManager(_:didChangeAuthorization:)` `CLLocationManagerDelegate` function to the new delegate function `locationManagerDidChangeAuthorization(_:)` in the now `BMLocationManager.swift`

https://github.com/user-attachments/assets/3c65ab79-93d8-443a-ac7d-4854822669c0


berkeley-mobile/Data/BMLocationManager.swift
```diff
@@ -0,0 +1,105 @@
...
+
+    func locationManager(_ manager: CLLocationManager, didUpdateLocations locations: [CLLocation]) {
+        if let location = locations.last {
+            userLocation = location
+            BMLocationManager.notificationCenter.post(name: .locationUpdated, object: location)
+        }
+    }
+
+    func locationManager(_ manager: CLLocationManager, didFailWithError error: Error) {
+        if let clErr = error as? CLError {
```
Chhumbucket, line 98: Only worried about the clErr being too similar to CLError, is it possible to specific the variable?

justinwongeecs: @jshaha Yeah that also could work. 

But I'm not quite sure about the similarity between the local variable `clError` and `CLError`? Here in we are optionally casting `error` into of type `CLError` so it makes sense to name the variable `clError` since is a `ClError`.

berkeley-mobile/Data/BMLocationManager.swift
```diff
@@ -0,0 +1,105 @@
...
+
+// MARK: CLLocationManagerDelegate
+
+extension BMLocationManager: CLLocationManagerDelegate {
+    func locationManagerDidChangeAuthorization(_ manager: CLLocationManager) {
+        handleLocationAuthorization(fromDelegate: true)
+    }
+
+    func locationManager(_ manager: CLLocationManager, didUpdateLocations locations: [CLLocation]) {
+        if let location = locations.last {
```
Chhumbucket, line 91: Rewrite the if let into a guard statement

berkeley-mobile/Data/BMLocationManager.swift
Chhumbucket, line 98: Rewrite as a guard statement

Chhumbucket (approved):

jshaha (approved):
Cute little update 👍

PR #347, opened 2025-03-20
BER-70: Rewrite Home Search View

baeuke:
## Updates  
- Removed old UIKit components:  
  - `SearchResultsView`
  - `SearchBarView`
  - `SearchResultCell` → `SearchResultsListRowView` (SwiftUI)
  - `SearchBarDelegate` implementation in `MapViewController`
  - `isShowingSearchResultsView` property in `MapViewController` (had no effect)  

- Added:  
  - `SearchBarView` (SwiftUI)
  - `SearchResultsView` (SwiftUI)
  - `BMContentUnavailableView` (custom `ContentUnavailableView`, temporary iOS 16 fix)  

- Refactored:  
  - Cleaned up `viewDidLoad()` of MapViewController
  - Consistent layout in `setupSubviews()` of MapViewController: now most components' constraints are set by NSLayoutConstraint.activate
  - Added .idle case in `SearchResultsState` enum for cases when search textfield is empty
  - Moved `UIStackView` extension from `SearchResultCell` (deleted) to `MapMarkerDetailView`  
  - Preserved `SearchResultsViewDelegate` as it is used by:  
    - `StudyViewController`  
    - `FitnessViewController`  
    - `DiningViewController`  

### Issue:  
To provide focus environment access for hosted SwiftUI views, `homeViewController` was introduced in `MainContainerViewController`:  
```swift
// Previous:
homeView = UIHostingController(rootView: HomeView()).view!  

// Now:
homeViewController = UIHostingController(rootView: HomeView())
homeView = homeViewController.view!
```

This caused extra top and bottom padding in the map view, affecting the search bar and detail cards.  

### Fix  
- Issue seems related to `UIHostingController` and `layoutMarginsGuide`.  
- Updated constraints:  
  - Used `view.bottomAnchor/topAnchor` instead of `layoutMarginsGuide` where needed.  
  - `leftAnchor` and `rightAnchor` with `layoutMarginsGuide` work fine.  

## Future Improvements  
- **Refactor to separate concerns:**  
  - Move `choosePlacemark` and `chooseMapMarker` to `SearchViewModel` while resolving dependencies on `MapViewController`.  

https://github.com/user-attachments/assets/d1bc4ea4-7c51-4ae5-b52d-db0a7453045b


Chhumbucket (commented):
I feel that the zoom in for the location is too fast

baeuke:
> I feel that the zoom in for the location is too fast
@Chhumbucket overall we use builtin setRegion of MapKit for zooming in. i added an extension with a custom duration; try now and let me know

berkeley-mobile/Map/SearchResultsListRowView.swift
```diff
@@ -0,0 +1,72 @@
...
+
+    var distance: Double? {
+        guard
+            let userLocation = LocationManager.shared.userLocation,
+            let placemarkLocation = placemark.location
+        else {
+            return nil
+        }
+        
+        return round(userLocation.distance(from: placemarkLocation) / 1600.0 * 10) / 10
```
justinwongeecs, line 23: Use `Measurement` to convert distance to miles (I'm assuming that you are converting it into miles)?
Checkout the `distanceToUser` computed property in `HasLocation.swift`.
Maybe you can abstract away this functionality as an extension on `Measurement`?

berkeley-mobile/Map/SearchResultsListRowView.swift
```diff
@@ -0,0 +1,72 @@
...
+    var body: some View {
+        HStack {
+            VStack(alignment: .center) {
+                Image(systemName: "mappin")
+                    .resizable()
+                    .scaledToFit()
+                    .frame(width: 8)
+                    .foregroundStyle(Color(BMColor.searchBarIconColor))
+                
+                if let distance { // If user location is unavailable, then we don't show the distance
```
justinwongeecs, line 35: The comment is redundant. It can be inferred from the context.

berkeley-mobile/Map/SearchResultsListRowView.swift
```diff
@@ -0,0 +1,72 @@
...
+            VStack(alignment: .leading) {
+                Text(placemark.searchName ?? "Not Available")
+                    .font(Font(BMFont.regular(16)))
+                
+                Text(placemark.locationName ?? "Not Available")
+                    .font(Font(BMFont.regular(14)))
+            }
+            .frame(height: 36)
+            
+            Spacer() // To push stuff to the leading edge.
```
justinwongeecs, line 51: Comment is redundant.

berkeley-mobile/Map/SearchResultsListRowView.swift
```diff
@@ -0,0 +1,72 @@
...
+        }
+        .padding(10)
+        .frame(height: 60.0)
+        .frame(maxWidth: .infinity)
```
justinwongeecs, line 55: Isn't this by default?

berkeley-mobile/Map/SearchResultsListRowView.swift
```diff
@@ -0,0 +1,72 @@
...
+    }
+}
+
+#Preview {
+    let previewCoordinates = CLLocation(latitude: 37.87538, longitude: -122.25612109999999)
+    let previewPlacemark = MapPlacemark(loc: previewCoordinates, name: "Foothill", locName: "2700 Hearst Ave, Berkeley, CA 94720")
+    
+    return VStack {
```
justinwongeecs, line 63: Don't need `return`?

berkeley-mobile/Map/SearchResultsView.swift
```diff
@@ -1,204 +1,73 @@
...
-    }
-    
-    func updateView() {
-        switch state {
-        case .error(let error):
-            if error.localizedDescription != "The operation couldn’t be completed. (MKErrorDomain error 4.)" {  // Temporary workaround
-                errorLabel.text = error.localizedDescription
+struct SearchResultsView: View {
+    @EnvironmentObject var viewModel: SearchViewModel
+    let spacerHeight: CGFloat = 128 // Pushing some content visually to the center due to the top anchor of the SearchResultsView starting from the bottom of the SearchBarView
```
justinwongeecs, line 14: Make this private?

berkeley-mobile/Map/SearchResultsView.swift
```diff
@@ -1,204 +1,73 @@
...
+    
+    var body: some View {
+        VStack {
+            switch viewModel.state {
+            case .idle:
+                BMContentUnavailableView(iconName: "text.magnifyingglass", title: "Type Something", subtitle: "Search for anything you need.")
+                Spacer()
+                    .frame(height: spacerHeight)
```
justinwongeecs, line 22:
```swift
Spacer()
    .frame(height: spacerHeight)` 
```

is used after every case. Perhaps you can move it outside the `switch` statement but inside the `VStack`?

berkeley-mobile/Map/SearchResultsView.swift
```diff
@@ -1,204 +1,73 @@
...
-    }
-    
+#Preview {
+    SearchResultsView()
+        .padding()
+        .environmentObject(SearchViewModel(chooseMapMarker: { mapMarker in
+            print("\(mapMarker)")
+        }, choosePlacemark: { placemark in
+            print("\(placemark)")
+        }))
```
justinwongeecs, line 72: Rewrite to

`.environmentObject(SearchViewModel(chooseMapMarker: { _ in }) { _ in }))`

Do the print statements add any additional value or information?

berkeley-mobile/Map/SearchViewModel.swift
```diff
@@ -0,0 +1,112 @@
...
+    func clearSearchText() {
+        searchText = ""
+    }
+    
+    func searchLocations(_ keyword: String, completion: (([MapPlacemark], Error?) -> Void)? = nil) {
+        
+        state = .loading
+        
+        DispatchQueue.global(qos: .userInteractive).async { [weak self] in
+            guard let self = self else { return }
```
justinwongeecs, line 64: Rewrite to 
```swift
guard let self else {
    return
}
```

berkeley-mobile/Map/SearchViewModel.swift
```diff
@@ -0,0 +1,112 @@
...
+        self.choosePlacemark = choosePlacemark
+    }
+    
+    // SearchBarView's funcs:
+    func clearSearchText() {
+        searchText = ""
+    }
+    
+    func searchLocations(_ keyword: String, completion: (([MapPlacemark], Error?) -> Void)? = nil) {
+
```
justinwongeecs, line 60: Remove newline

berkeley-mobile/Map/SearchViewModel.swift
```diff
@@ -0,0 +1,112 @@
...
+    @Published var state = SearchResultsState.idle
+    var chooseMapMarker: (MapMarker) -> Void
+    var choosePlacemark: (MapPlacemark) -> Void
+    
+    init(chooseMapMarker: @escaping (MapMarker) -> Void, choosePlacemark: @escaping (MapPlacemark) -> Void) {
+        self.chooseMapMarker = chooseMapMarker
+        self.choosePlacemark = choosePlacemark
+    }
+    
+    // SearchBarView's funcs:
```
justinwongeecs, line 54: If you want to keep these comments, I think it would be better to make them as MARKs instead as it helps delinate or separate these functions more clearly in the text editor.

berkeley-mobile/Map/SearchViewModel.swift
```diff
@@ -0,0 +1,112 @@
...
+            DispatchQueue.main.async {
+                self.updatePlacemarksList(newPlacemarks: placemarks, error: nil)
+                completion?(placemarks, nil)
+            }
+        }
+    }
+    
+    // SearchResultsView's funcs:
+    func updatePlacemarksList(newPlacemarks: [MapPlacemark]?, error: Error?) {
+        if let error = error {
```
justinwongeecs, line 84: `let error = error` &rarr; `let error`

berkeley-mobile/Map/SearchViewModel.swift
```diff
@@ -0,0 +1,112 @@
...
+            state = .error(error)
+            return
+        }
+        
+        guard let newPlacemarks = newPlacemarks, !newPlacemarks.isEmpty else {
```
justinwongeecs, line 89: `let newPlacemarks = newPlacemarks` &rarr; `let newPlacemarks`

berkeley-mobile/Map/SearchViewModel.swift
```diff
@@ -0,0 +1,112 @@
...
+        }
+        
+        state = .populated(newPlacemarks)
+    }
+    
+    func selectListRow(_ placemark: MapPlacemark) {
+        searchText = ""
+        isSearchBarFocused = false
+        
+        if let placemark = placemark.item as? MapMarker {
```
justinwongeecs, line 105: I think it would be suitable to rename the local variable `placemark` to `mapMarker` or `marker`  to be more specific that it has been casted to a `MapMarker`.

berkeley-mobile/Utils/View+Extension.swift
```diff
@@ -42,5 +49,5 @@ struct Cardify: ViewModifier {
 extension View {
     func cardify() -> some View {
         self.modifier(Cardify())
-    }
+        }
```
justinwongeecs, line 52: Remove extra space?

justinwongeecs (changes requested):

Chhumbucket (commented):
There should be some padding between the searchbar and the items that pop up on the search. These two items feel too close together. I linked an image below. 
Screenshot 2025-03-20 at 7 48 26 PM

berkeley-mobile/Map/BMContentUnavailableView.swift
```diff
@@ -0,0 +1,40 @@
...
+
+import SwiftUI
+
+struct BMContentUnavailableView: View {
+    let iconName: String
+    let title: String
+    let subtitle: String
+    
+    var body: some View {
+        VStack(spacing: 0) {
```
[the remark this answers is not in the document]
baeuke: sure! but it seems by default VStack actually has a little spacing

berkeley-mobile/Map/MapMarkerDetailView.swift
```diff
@@ -275,3 +275,17 @@ enum MapMarkerDetail {
     }
     
 }
+
+// MARK: - UIStackView
+extension UIStackView {
+    convenience init(arrangedSubviews: [UIView]? = nil, axis: NSLayoutConstraint.Axis, distribution: UIStackView.Distribution, spacing: CGFloat) {
+        if let arrangedSubviews = arrangedSubviews {
```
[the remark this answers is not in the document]
baeuke: good catch! this was written before i just moved it from the deleted file but yeah need to refactor

berkeley-mobile/Map/MapViewController.swift
```diff
@@ -150,6 +139,48 @@ class MapViewController: UIViewController, SearchDrawerViewDelegate {
...
     }
+
+    private func createMaskView() {
+        maskView = UIHostingController(rootView: MaskView()).view
+        maskView.translatesAutoresizingMaskIntoConstraints = false
+        maskView.backgroundColor = UIColor.clear
+    }
+    
+    private func createSearchBarComponents() {
+        // SearchViewModel
```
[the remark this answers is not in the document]
baeuke: sure thing

berkeley-mobile/Map/MapViewController.swift
```diff
@@ -560,3 +541,11 @@ extension MapViewController {
         searchAnnotation = nil
     }
 }
+
+extension MKMapView {
+    func setRegionWithDuration(_ zoomRegion: MKCoordinateRegion, duration: TimeInterval) {
+        MKMapView.animate(withDuration: duration, delay: 0, usingSpringWithDamping: 0, initialSpringVelocity: 10, options: UIView.AnimationOptions.curveEaseIn, animations: {
```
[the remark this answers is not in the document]
baeuke: oh i just found this workaround from stack overflow, mostly to experience slower animation, did you try running on simulator or on your phone? how does it feel?
+) is it better to use UIView.animate? just MKMapView extends UIView right, so maybe it has specific overrides for animate?

berkeley-mobile/Map/SearchBarView.swift
```diff
@@ -1,176 +1,97 @@
...
-        if !isSearching {
-            textField.resignFirstResponder()
+        .padding()
+        .background(.regularMaterial)
+        .clipShape(.rect(cornerRadius: 15))
+        .shadow(color: .black.opacity(0.3), radius: 8)
+        .onChange(of: isFocused) { newValue in // onChange syntax will need to change in later iOS
+            if viewModel.isSearchBarFocused != newValue {
+                viewModel.isSearchBarFocused = newValue
+            }
```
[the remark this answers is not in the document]
baeuke: the thing is we kinda have this binding: if viewModel.isSearchBarFocused changes externally then we change isFocused, if isFocused is changed internally then it propogates to viewModel.isSearchBarFocused - so we basically have to .onChange modifiers and if we remove `if viewModel.isSearchBarFocused != newValue ` part then if say viewModel.isSearchBarFocused chnages then it will set isFocused, but change of isFocused will trigger a change of viewModel.isSearchBarFocused and so on

berkeley-mobile/Map/SearchBarView.swift
```diff
@@ -1,176 +1,97 @@
...
+            
+            if newValue == true {
+                onSearchBarTap?(true)
+            } else {
+                onSearchBarTap?(false)
+            }
```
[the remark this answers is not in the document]
baeuke: nice catch!

berkeley-mobile/Map/SearchBarView.swift
```diff
@@ -1,176 +1,97 @@
...
-        textField.text = textField.text?.trimmingCharacters(in: .whitespaces).count == 0 ? "" : textField.text
-        setButtonStates(hasInput: textField.text?.count != 0, isSearching: true)
-        
-        startSearch = DispatchWorkItem { [weak self] in
-            guard let self = self, let startSearch = self.startSearch, !startSearch.isCancelled else { return }
-            self.delegate?.searchbarTextDidChange(textField)
+        .onChange(of: viewModel.isSearchBarFocused) { newValue in
+            if isFocused != newValue {
+                isFocused = newValue
+            }
```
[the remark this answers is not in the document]
baeuke: same as with the comment above, you can try to put prints within onchange blocks and tap on the search bar and see what i mean

berkeley-mobile/Map/SearchBarView.swift
```diff
@@ -1,176 +1,97 @@
...
-        setButtonStates(hasInput: textField.text?.count != 0, isSearching: true)
-        delegate?.searchbarTextFieldDidBeginEditing(textField)
     }
     
-    func textFieldDidEndEditing(_ textField: UITextField) {
-        textField.text = textField.text?.trimmingCharacters(in: .whitespaces).count == 0 ? "" : textField.text
-        setButtonStates(hasInput: textField.text?.count != 0, isSearching: false)
-        delegate?.searchbarTextFieldDidEndEditing(textField)
-        textField.resignFirstResponder()
+    private var leftIcon: some View {
```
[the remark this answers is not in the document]
baeuke: how about searchOrBackIcon ?

berkeley-mobile/Map/SearchBarView.swift
```diff
@@ -1,176 +1,97 @@
...
+                .fontWeight(.semibold)
+                .frame(width: 20, alignment: .center) // So that changing an icon didn't move the TextField
+        }
     }
     
-    func textFieldShouldReturn(_ textField: UITextField) -> Bool {
-        textField.text = textField.text?.trimmingCharacters(in: .whitespaces).count == 0 ? "" : textField.text
-        setButtonStates(hasInput: textField.text?.count != 0, isSearching: false)
-        return delegate?.searchbarTextShouldReturn(textField) ?? true
+    private var rightIcon: some View {
```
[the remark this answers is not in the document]
baeuke: clearTextIcon?

berkeley-mobile/Map/SearchBarView.swift
```diff
@@ -1,176 +1,97 @@
...
-    func searchbarTextShouldReturn(_ textField: UITextField) -> Bool
+#Preview {
+    SearchBarView()
+        .padding()
+        .ignoresSafeArea()
+        .environmentObject(SearchViewModel(chooseMapMarker: { mapMarker in
+            print("\(mapMarker)")
+        }, choosePlacemark: { placemark in
+            print("\(placemark)")
+        }))
```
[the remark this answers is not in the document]
baeuke: will do! noo they were there cuz i thought we need some functions for preview to work😂

berkeley-mobile/Map/SearchBarView.swift
```diff
@@ -1,176 +1,97 @@
...
-        delegate?.searchbarTextFieldDidEndEditing(textField)
-        textField.resignFirstResponder()
+    private var leftIcon: some View {
+        Button(action: {
+            if isFocused {
+                isFocused = false
+                viewModel.clearSearchText()
+            } else {
+                isFocused = true
+            }
```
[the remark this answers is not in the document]
baeuke: niice

berkeley-mobile/Map/SearchResultsListRowView.swift
baeuke: removed .frame(maxWidth: .infinity) 👍🏻

berkeley-mobile/Map/SearchResultsListRowView.swift
baeuke: oh wow i thought if we have some other stuff like var inits etc then we need a return, kinda like viewbuilder

berkeley-mobile/Map/SearchResultsView.swift
baeuke: spacerHeight?

berkeley-mobile/Map/SearchResultsView.swift
baeuke: it's not used for .populated case that's the thing, and i thought having another if block for checking state would be wrong when we have switch already TT

berkeley-mobile/Map/SearchResultsView.swift
baeuke: good

berkeley-mobile/Map/SearchViewModel.swift
baeuke: is it okay to have nested MARKs?

berkeley-mobile/Map/SearchViewModel.swift
baeuke: fixed👍🏻

baeuke:
> There should be some padding between the searchbar and the items that pop up on the search. These two items feel too close together. I linked an image below. Screenshot 2025-03-20 at 7 48 26 PM

nice find! added spacing of 10

berkeley-mobile/Map/MaskView.swift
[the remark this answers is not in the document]
baeuke: So, this was sort of a temporary fix. but i got an idea, will test and let you know

berkeley-mobile/Map/SearchResultsListRowView.swift
baeuke: Yeah good idea! since we have the mileage showing in many places. should i add the extension to View+Extension file?

baeuke: Also, I don't think we need rounding right (16575.585.. -> 16576.0)? since we specify the format in the Text view anyways, and it would be more accurate without rounding i guess

baeuke: Okay @justinwongeecs there was a weird thing with extending Measurement: 1) we'd have to create an instance of Measurement every time in the code or 2) we'd have to make it static BUT write Measurement<UnitLenght>.funcName etc. So instead i just extended CLLocation, seems cleaner now. you can check at the end of file `SearchResultsListRowView.swift`

baeuke: + cleaned up `distanceToUser` in `HasLocation.swift` to use the new `distanceFromUser()`.
Double-checked the values, everything is correct 👍🏻

berkeley-mobile/Map/MaskView.swift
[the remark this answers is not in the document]
baeuke: okay so i tried doing this (kinda the way it "should have been"):

Screenshot 2025-03-20 at 11 36 55 PM

so all: mask view, search bar, and searchResults are one "SuperSearchBar" which is mainly a ZStack. 
But, no matter the hitTesting combinations and various backgrounds set to .clear, it's placed on top of the map and is not pass through in the empty area so we can't interact with the map:

Screenshot 2025-03-20 at 11 33 08 PM

There is an ugly solution that should work though: 
is leaving the SearchBar as it is, but putting searchResults into a ZStack and rather than having it placed under searchBar we make it full screen view but adjust the placement of the SerachResultsComponents how it should be. then it will be quite hardcoded layout within the swiftUI file so in anyway we end up with a temporary fix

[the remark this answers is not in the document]
baeuke: okayxD i decided to go with the ugly solution since it's more long term in terms of moving to swiftui and it will be easier to further migrate. no more MaskView, it's part of the SearchResultsView, but SearchResultsView is separate from SearchBar for now. I think once the whole thing will be one big ZStack then we can combine SearchResultsView and SearchBar together.

berkeley-mobile/Map/SearchViewModel.swift
baeuke: I meant i thought we use marks for larger separation like classes structs etc. is it okay to use within the class

berkeley-mobile/Map/SearchBarView.swift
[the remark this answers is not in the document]
justinwongeecs: So if I understand your comment, you're saying that it will cause an infinite loop? But it doesn't as `onChange` only gets triggered if the value is set to something that differs from its original value. So if `viewModel.isSearchBarFocused` changes it will set `isFocused` but then it will stop there since when `isFocused` sets `viewModel.isSearchBarFocused` the `onChange` closure wouldn't get called anymore thus ending the loop. I added print statements for each circumstance, and I got the same behavior.

berkeley-mobile/Map/MapViewController.swift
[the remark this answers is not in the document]
justinwongeecs: A subclass can't override a static/class method of its superclass directly. It has to declare a new method of the same signature. But on the documentation, `MKMapView` doesn't have an `animate(...)` function so thus we know that it comes from `UIView`. I think it's better to use `UIView.animate` for more clarity and convention.

justinwongeecs (changes requested):

justinwongeecs (changes requested):

berkeley-mobile/Map/SearchResultsView.swift
```diff
@@ -1,204 +1,70 @@
...
+                ForEach(viewModel.state.placemarks, id: \.location) { placemark in
+                    Button(action: {
+                        viewModel.selectListRow(placemark)
+                    }) {
+                        SearchResultsListRowView(placemark: placemark)
+                    }
+                    .buttonStyle(SearchResultsListRowButtonStyle())
+                }
+            }
         }
```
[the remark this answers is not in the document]
baeuke: nice catch!

justinwongeecs (changes requested):
https://github.com/user-attachments/assets/c2815932-3d1e-4ae2-bda5-b4f2013a89c2

Perhaps when scrolling through the search results, we dismiss the keyboard so that the bottom entries are viewable? Because on an actual iPhone there's no way to dismiss it sort of what Instagram does when you are searching something in the search UI.

The focused state with the search bar is kinda of tricky because the logic is intertwined with the display of the mask view.

justinwongeecs (approved):
Looks good!

Chhumbucket (approved):

PR #348, opened 2025-03-24
BER-75: Fix MapMarkersDropdownView Regression: Row Tap Target and Font Weight

iychi:
### MapMarkersDropdownView:
  Added a `Spacer()` to the end of the `HStack` to expand the tap target area
  
### Font Weight
BMFont does not have a `semibold` variant, so I used `medium`.  
I've attached screenshots comparing `regular` vs. `medium` — feel free to share which looks better!

### regular
BMFont regular

### medium
BMFont medium

### DIFF
--- a/berkeley-mobile/Map/MapMarkersDropdownView.swift
+++ b/berkeley-mobile/Map/MapMarkersDropdownView.swift
@@ -65,7 +65,7 @@ struct MapMarkersDropdownView: View {
                     HStack(spacing: 15) {
                         Image(uiImage: type.icon())
                         Text(type.rawValue)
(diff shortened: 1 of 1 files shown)


justinwongeecs (commented):
I actually like the regular font! It's similar to what Apple has in Messages. Any thoughts?

jshaha (approved):
I like the regular font as well. Looks good :)

iychi:
I also like the regular one. It looks just right—not too heavy.

justinwongeecs:
@iychi Do you want to change the code back to regular. I believe right now it's medium?

jshaha (approved):
Looks good :)

PR #351, opened 2025-03-27
BER-89: Fix Incorrect StateObject Configuration For HomeViewModel

justinwongeecs:
The argument into `StateObject` has a `@autoclosure` applied to it. What this does is that it wraps the argument in a closure and ensures that it only gets initialized once during its lifetime. However, since we create an intermediate variable to hold `HomeViewModel`, each time `HomeView`'s `init()` gets called, an unnecessary instantiation of `HomeViewModel` occurs and then immediately gets released back into memory.


jshaha (dismissed):
Looks good! 😁

YHC66:
Looks good!

YHC66 (approved):

jshaha (approved):
looking good again. :)

PR #355, opened 2025-04-03
BER-91: Stale Data With Gym Occupancy Views

justinwongeecs:
After some time, the gym occupancy views do not update/refresh after `FitnessViewController` is out of view 
- Removed `onDisappear` view modifier since that stopped the updating


hetvi15986 (commented):
looking good

iychi (approved):

PR #356, opened 2025-04-03
Loading state to ResourcesViewModel

ishatailor:


ishatailor:
wait could i get reviews on this it's so quick i promise

justinwongeecs (changes requested):
There's more than +4mil line changes which is definitely not necessary

PR #361, opened 2025-04-15
BER-92: Gym Occupancy Home Screen Widgets

justinwongeecs:
- Added new `.systemSmall` widget that displays both the gym occupancies for both RSF Weight Rooms and CMS Fitness Center
- Going the the Fitness Tab in the app will automatically refresh widgets 
- Refreshes ~every 15 mins (fastest that we can do as we are restricted by WidgetKit and iOS)

screenshot_2025-04-14_at_5 37 14___pm_480
screenshot_2025-04-14_at_5 37 37___pm_480

### DIFF
--- a/BerkeleyMobileWidget/GymOccupancyWidget.swift
+++ b/BerkeleyMobileWidget/GymOccupancyWidget.swift
@@ -1,0 +1,209 @@
+//
+//  GymOccupancyWidget.swift
+//  GymOccupancyWidget
+//
+//  Created by Justin Wong on 4/13/25.
+//  Copyright © 2025 ASUC OCTO. All rights reserved.
+//
+
+import SwiftUI
+import WidgetKit
+
+typealias GymOccupancyChange = (prior: Double?, current: Double)
+
+struct GymOccupancyEntry: TimelineEntry {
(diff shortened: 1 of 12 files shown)


BerkeleyMobileWidget/GymOccupancyWidget.swift
```diff
@@ -0,0 +1,212 @@
...
+
+typealias GymOccupancyChange = (prior: Double?, current: Double)
+
+struct GymOccupancyEntry: TimelineEntry {
+    static let defaultRSFOccupancyPercentages: GymOccupancyChange = (100, 91)
+    static let defaultStadiumOccupancyPercentages: GymOccupancyChange = (43, 68)
+    
+    let date: Date
+    let RSFOccupancyPercentages: GymOccupancyChange
+    let stadiumOccupancyPercentages: GymOccupancyChange
```
baeuke, line 20: maybe change the variable name from `RSFOccupancyPercentages` to `rsfOccupancyPercentages` ? (it'd also make it consistent with the viewmodel name)
or is tuple typealias is kinda like a whole type? maybe then we should use upper case for `stadiumOccupancyPercentages`

baeuke (changes requested):

PR #364, opened 2025-04-23
BER-102: Create Recent-Search History Mechanism and UI For Search Bar

baeuke:
- created a new light `CodableMapPlacemark` type to be able to save `MapPlacemarks` with `UserDefaults`
- extended `MapPlacemark` and added functions to interop with `CodableMapPlacemark`
- added `RecentSearchManager` layer between `SearchViewModel` and `UserDefaults` that adds, deletes, loads, and saves recent searches
- refactored existing `struct UserDefaultKeys` and added  `UserDefaults` extension for easier use of a new `UserDefaultsKeys` enum

https://github.com/user-attachments/assets/df804b9c-fcae-4406-866b-d16dccec2095


berkeley-mobile/Map/MapPlacemark.swift
```diff
@@ -27,5 +28,52 @@ class MapPlacemark {
...
+        )
+    }
     
+    static func fromCodable(_ codablePlacemark: CodableMapPlacemark) -> MapPlacemark? {
+        let delta = 0.0001 // threshold check in case of a precision loss when encoding/decoding
+        
+        guard let searchItem = DataManager.shared.searchable.first(where: {
+            $0.searchName == codablePlacemark.searchName &&
+            abs($0.location.0 - codablePlacemark.latitude) < delta &&
+            abs($0.location.1 - codablePlacemark.longitude) < delta
```
justinwongeecs, line 65: Do we need the precision check for the latitude and longitude with delta?

berkeley-mobile/Map/RecentSearchManager.swift
```diff
@@ -0,0 +1,59 @@
...
+    
+    static func add(_ placemark: MapPlacemark, to recentSearch: [CodableMapPlacemark]) -> [CodableMapPlacemark] {
+        guard let codablePlacemark = placemark.toCodable() else {
+            return recentSearch
+        }
+        
+        var updatedRecentSearch = recentSearch
+        
+        // remove if already exists
+        if let index = updatedRecentSearch.firstIndex(where: { $0 == codablePlacemark }) {
```
justinwongeecs, line 31: Remove comment. Rename `index` to `removeIndex` or something along the lines of that

berkeley-mobile/Map/SearchResultsListRowView.swift
```diff
@@ -52,7 +54,33 @@ struct SearchResultsListRowView: View {
...
+    var codablePlacemark: CodableMapPlacemark
+    
+    var body: some View {
+        HStack {
+            Image(systemName: "clock.arrow.trianglehead.counterclockwise.rotate.90")
+                .resizable()
+                .scaledToFit()
+                .frame(width: 24)
+                .foregroundStyle(Color(BMColor.searchBarIconColor))
+                .frame(width: 52)
```
justinwongeecs, line 70: Are the two `.frame(width:)` view modifiers necessary?

berkeley-mobile/Map/SearchResultsView.swift
```diff
@@ -20,12 +20,17 @@ struct SearchResultsView: View {
...
                 case .idle:
-                    BMContentUnavailableView(iconName: "text.magnifyingglass", title: "Type Something", subtitle: "Search for anything you need.")
-                    
+                    if viewModel.recentSearch.isEmpty {
+                        BMContentUnavailableView(iconName: "text.magnifyingglass", title: "Type Something", subtitle: "Search for anything you need.")
+                    } else {
+                        recentSearchList
+                            .padding(.top, searchResultsListTopPadding)
+                            .padding(.horizontal, 21)
+                            .padding(.bottom, 96)
```
justinwongeecs, line 32: Maybe use the other padding view modifier function signature? 

```swift
recentSearchList
    .padding(EdgeInsets(top: searchResultsListTopPadding, leading: 21, bottom: 96, trailing: 21))```

File: berkeley-mobile/Map/SearchResultsView.swift [hunk 5]
```diff
@@ -64,6 +69,45 @@ struct SearchResultsView: View {
...
+        VStack {
+            HStack {
+                Text("Recents")
+                    .font(Font(BMFont.regular(14)))
+                
+                Spacer()
+                
+                Button(action: {
+                    viewModel.deleteAllRecentSearch()
+                }) {
```
justinwongeecs, line 83: Perhaps try experimenting with adding an explicit animation with `withAnimation { ... }`? or an implicit animation on `viewModel.recentSearch`

berkeley-mobile/Map/SearchViewModel.swift
```diff
@@ -43,12 +43,15 @@ class SearchViewModel: ObservableObject {
     @Published var searchText = ""
     @Published var isSearching = false
     @Published var state = SearchResultsState.idle
+    @Published var recentSearch: [CodableMapPlacemark]
```
justinwongeecs, line 46: Rename to `recentSearches`?

justinwongeecs (changes requested):

berkeley-mobile/Map/MapPlacemark.swift
baeuke: I think so. It seems that encoding/decoding step might lead to loss in precision. We use JSONEncoder/Decoder when saving to (fetching from) UserDefaults and then compare to try to find original value among searchables, and we don't use values of latitude and longitude up to a certain floating point, we use full values;
When i checked equating == and printing values for several examples it worked, but i think better to be safe since if they don't match then it might be hard to find where the issues comes from later on.

berkeley-mobile/Map/SearchResultsListRowView.swift
baeuke: yes, so the first one .frame(width: 24) is for setting/controlling the size of the system image.
and second .frame(width: 52) is setting the frame for the whole image view, so that they weren't too close with title text

berkeley-mobile/Utils/UserDefaults+Extension.swift
```diff
@@ -0,0 +1,30 @@
...
+//  Copyright © 2025 ASUC OCTO. All rights reserved.
+//
+
+import Foundation
+
+enum UserDefaultsKeys: String {
+    case numAppLaunchForFeedbackForm = "numAppLaunchForFeedbackForm"
+    case academicEventsLastSavedDate = "academicEventsLastSavedDate"
+    case campuswideEventsLastSavedDate = "campuswideEventsLastSavedDate"
+    case recentSearches = "RecentSearch"
```
justinwongeecs, line 15: `"RecentSearch` &rarr `"recentSearches"`

justinwongeecs (changes requested):

justinwongeecs (approved):
👍

PR #433, opened 2025-10-06
Display date and time separately in event detail view

Devansh-Ojha:
### Changes
- Separated event date and time into two lines for clarity
- Added SF Symbols:
  -  calendar for the date
  -  clock for the time
- added BMEventCalendarEntry.dateString  and BMEventCalendarEntry.timeString 
- Also shows tdy/tmr instead of exact date if event is tdy or tmr

### DIFF
--- a/berkeley-mobile/Events/EventDetailView.swift
+++ b/berkeley-mobile/Events/EventDetailView.swift
@@ -148,7 +148,7 @@ struct BMDetailHeaderView: View {
     @ViewBuilder
     private var timeView: some View {
         if let timePart = event.dateString.components(separatedBy: " / ").last {
-             EventDetailRow(systemImageName: "calendar", text: timePart)
+             EventDetailRow(systemImageName: "clock", text: timePart)
         }
     }
 


berkeley-mobile/Events/EventDataSource/BMEventCalendarEntry.swift
```diff
@@ -122,9 +122,11 @@ class BMEventCalendarEntry: NSObject, NSCoding, Identifiable, BMCalendarEvent, H
         
         self.type = coder.decodeObject(forKey: ArgumentNames.type) as? String
     }
+
```
justinwongeecs: Remove newline

berkeley-mobile/Events/EventDetailView.swift
```diff
@@ -73,8 +73,9 @@ struct EventDetailView: View {
     
     private func presentLinkAlert(type: AlertType?) {
         guard let type else {
-            return
-        }
+                    return
+                }
+
```
justinwongeecs: the extra indentations look a bit weird 

should be indented like this:

```swift
guard let type else {
    return
}
```

berkeley-mobile/Events/EventDetailView.swift
```diff
@@ -157,6 +162,19 @@ struct BMDetailHeaderView: View {
     }
 }
 
+struct EventDetailRow: View {
```
justinwongeecs: To better find different struct views in the mini bar, we usually add a MARK statement above the struct declaration. 

For consistency throughout the codebase, there should always be 2 new lines above a MARK and one new line after a MARK. The only exception is that if you have a MARK after the imports at the top of a file, there should only be 1 new line before it. 

Checkout of how we use the MARKs in the different places in the file.

Devansh-Ojha:
Fixed the issue regarding MARk and newline spaces

berkeley-mobile/Events/EventDetailView.swift
```diff
@@ -97,7 +97,6 @@ struct EventDetailView: View {
     }
 }
 
-
 // MARK: - BMDetailHeaderView
```
justinwongeecs: there should be **2** newlines before the MARK

berkeley-mobile/Events/EventDetailView.swift
```diff
@@ -133,14 +138,14 @@ struct BMDetailHeaderView: View {
         }
     }
     
+    private var dateView: some View {
+        let datePart = event.dateString.components(separatedBy: " / ").first ?? ""
+        return EventDetailRow(systemImageName: "calendar", text: datePart)
```
justinwongeecs: We shouldn't display the dateView if `datePart` is `""`. 

so instead we should do something like this: 

```swift
@ViewBuilder
private var dateView: some View {
    if let dateString = event.dateString.components(separatedBy: " / ").first {
         return EventDetailRow(systemImageName: "calendar", text: dateString)
    }
}
```

justinwongeecs (changes requested):

justinwongeecs (changes requested):

PR #453, opened 2025-10-20
Open Period Info Button

Chhumbucket:


berkeley-mobile/Home/Dining/DiningDetailView.swift
```diff
@@ -104,10 +105,17 @@ struct DiningAllDayView: View {
 // MARK: - DiningItemsListView
 
 struct DiningItemsListView: View {
+    @Binding var selectedTabIndex: Int
     var categoriesAndMenuItems: [BMMealCategory]
+    var diningHall: BMDiningHall
     
     var body: some View {
         VStack(spacing: 20) {
+            if selectedTabIndex < diningHall._hours.count {
```
justinwongeecs: what happens if there is only one element in `diningHall._hours` and let's say it's for lunch. the `selectedTabIndex` would be `1` which makes the condition false. 

I think once Zaid's PR goes through we can then index using the `selectedTabIndex` directly

PR #456, opened 2025-10-28
added Hiding Dining Details if location not open

ZaidMadanat:
Created visibility filter to only show meals if at least one menu item exists. Filter through `VisibleTabs` to remove the Breakfast/Dinner if they are empty.


justinwongeecs (changes requested):
I was thinking of not modifying `BMSegmentedControlView.swift` but instead in `DiningDetailView.swift` we dynamically pass in the filtered array of strings into the `tabNames` parameter for the `BMSegmentedControlView`. 

```swift
BMSegmentedControlView(tabNames: ["Breakfast", "Lunch", "Dinner"], selectedTabIndex: $selectedTabIndex)
```

Instead of hardcoding the three tabs, we should derive it dynamically within `DiningDetailView`.

PR #458, opened 2025-10-28
Guides

justinwongeecs:
Guides is one of the principal features on our roadmap this semester. Guides in Apple Maps, allow users to discover curated groups of places of interest. We think bringing a similar feature into Berkeley Mobile would be very beneficial for freshman and transfer students. 

For example, we could have a curated guide on local coffee shops, study spaces, places to escape to nature, essential eats, etc... 

Perhaps a very straightforwards place for Guides UI wise would be to have its dedicated tab in the Home Drawer along with "Dining", "Fitness", and "Study". All data for Guides will be stored in its own collection in Firestore. A list to display all of the guides and upon tap the detail view appears either via a sheet, in the home drawer, or a full screen overlay.

### DIFF
--- a/berkeley-mobile/Home/Guides/Guide.swift
+++ b/berkeley-mobile/Home/Guides/Guide.swift
@@ -42,5 +42,6 @@ struct GuidePlace: Codable, Identifiable, HasImage, HasLocation, HasPhoneNumber,
         case address
         case latitude
         case longitude
+        case websiteURLString
     }
 }


berkeley-mobile/Home/Guides/Guide.swift
```diff
@@ -0,0 +1,34 @@
...
+
+struct Guide: Codable, Identifiable {
+    @DocumentID var id: String?
+    var name: String
+    var description: String
+    var curator: String?
+    var places: [GuidePlace]
+}
+
+struct GuidePlace: Codable, HasImage, HasLocation, HasPhoneNumber, HasWebsite {
```
Chhumbucket: Maybe get GuidePlace to need an identifiable and filter with id

berkeley-mobile/Home/Guides/GuideDetailView.swift
```diff
@@ -0,0 +1,121 @@
...
+
+import SwiftUI
+
+struct GuideDetailView: View {
+    @Environment(GuidesViewModel.self) private var viewModel
+    
+    var guide: Guide
+    
+    var body: some View {
+        List(guide.places, id: \.name) { place in
```
Chhumbucket: If a place has a the same name as another place, would it be possible for the list to not consider that place, because it is using the name, I believe the other place will not appear.

PR #467, opened 2025-11-03
Feature (Home): Sort Section List Items by Distance to User

### LINKED ISSUE #436: Sort BMHomeSectionListView Items By User Location
Currently in the Dining, Fitness, and Study tabs of the Home Drawer, the items in `BMHomeSectionListView` are not ordered ascending by user location distance. 

We should order the items by user location distance ascending. So the first item is the closest to the user and the last item is the furthest away.

ananyadua27:
Changes Made
- Added a property sortedItems to BMHomeSectionListView
- Replaced items → sortedItems in both ForEach loops to reflect new ordering
- Reused existing distanceToUser for distance calculation instead of adding duplicate logic

Testing steps
- Verified on simulator: distances display correctly under each list item, items appear closest → farthest from the user
- No unintended impact on tapping, navigation, or visual styling

Resolves #436

### DIFF
--- a/berkeley-mobile/Home/BMHomeSectionListView.swift
+++ b/berkeley-mobile/Home/BMHomeSectionListView.swift
@@ -68,10 +68,10 @@ struct BMHomeSectionListView: View {
             List {
                 Section {
                     ForEach(sortedItems, id: \.name) { item in
-                        Button {
+                        Button(action: {
                             mapViewController.choosePlacemark(item: item)
                             selectionHandler?(item)
-                        } label: {
+                        }) {
                             HomeSectionListRowView(rowItem: item)
                                 .frame(width: 290)
                         }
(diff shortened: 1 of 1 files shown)


berkeley-mobile/Home/BMHomeSectionListView.swift
```diff
@@ -46,16 +58,18 @@ struct BMHomeSectionListView: View {
...
             List {
                 Section {
-                    ForEach(items, id: \.name) { item in
-                        Button(action: {
+                    ForEach(sortedItems, id: \.name) { item in
+                        Button {
                             mapViewController.choosePlacemark(item: item)
                             selectionHandler?(item)
-                        }) {
+                        } label: {
```
justinwongeecs: for buttons, we typically do something like this: 
```swift
Button (action: {
}) {

}
```
What you have is valid, but since label is the last parameter and it is a closure, we can use the trailing closure syntax: [https://www.hackingwithswift.com/sixty/6/5/trailing-closure-syntax](url)

PR #469, opened 2025-11-06
Empty view for BMHomeSection

Devansh-Ojha:
- Added an if condition to make sure that if there is nothing, we show a text instead of a blank screen. 
- Had to adjust the other if conditions a bit to accommodate that.

### DIFF
--- a/berkeley-mobile/Home/BMHomeSectionListView.swift
+++ b/berkeley-mobile/Home/BMHomeSectionListView.swift
@@ -16,22 +16,30 @@ struct BMHomeSectionListView: View {
     var selectionHandler: ((SearchItem & HasLocation & HasImage) -> Void)?
     
     var body: some View {
-        VStack {
-            if #unavailable(iOS 26.0) {
-                sectionHeaderView
-            }
-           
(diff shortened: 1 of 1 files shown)


Chhumbucket (dismissed):

berkeley-mobile/Home/BMHomeSectionListView.swift
```diff
@@ -49,26 +49,32 @@ struct BMHomeSectionListView: View {
...
-                        Button(action: {
-                            mapViewController.choosePlacemark(item: item)
-                            selectionHandler?(item)
-                        }) {
-                            HomeSectionListRowView(rowItem: item)
-                                .frame(width: 290)
+            if items.isEmpty {
+                Text("No Available Items")
+                    .frame(maxWidth: .infinity, maxHeight: .infinity)
+                    .padding()
```
justinwongeecs: We should use a `BMFont` for the text like this in `GuidesView.swift`: 

```swift
 Text("No Guides Available")
     .font(Font(BMFont.bold(20)))
     .foregroundStyle(.gray)
```

PR #475, opened 2025-11-18
Remove iOS 17 availability checks and keep else case

jayana-cpc:


berkeley-mobile/Safety/SafetyMapView.swift
```diff
@@ -18,16 +18,9 @@ struct SafetyMapView: View {
...
     
     var body: some View {
-        if #available(iOS 17.0, *) {
-            SafetyNewMapView(selectedSafetyLog: $selectedSafetyLog,
-                             isShowingLegend: $isShowingLegend,
-                             drawerViewState: $drawerViewState,
-                             isPresentingDetailView: isPresentingDetailView)
-        } else {
-            oldMapView
-        }
```
justinwongeecs: same thing here

justinwongeecs (changes requested):

justinwongeecs (changes requested):

PR #477, opened 2025-11-18
Added feedback button to search bar

jayana-cpc:


berkeley-mobile/Home/Search/SearchBarView.swift
```diff
@@ -26,8 +27,26 @@ struct SearchBarView: View {
...
+            if viewModel.searchText.isEmpty {
+                Button(action: {
+                    isPresentingFeedbackForm = true
+                }) {
+                    Image(systemName: "bubble.left.and.bubble.right")
+                        .foregroundStyle(Color(BMColor.searchBarIconColor))
+                        .fontWeight(.semibold)
+                        .frame(width: 30, alignment: .center)
+                }
+                .accessibilityLabel("Send Feedback")
```
justinwongeecs: refactor into its own private var

berkeley-mobile/Home/Search/SearchBarView.swift
```diff
@@ -26,8 +27,26 @@ struct SearchBarView: View {
...
+            } else {
+                Button(action: {
+                    viewModel.clearSearchText()
+                }) {
+                    Image(systemName: "xmark")
+                        .foregroundStyle(Color(BMColor.searchBarIconColor))
+                        .fontWeight(.semibold)
+                        .frame(width: 30, alignment: .center)
+                }
+                .accessibilityLabel("Clear search text")
```
justinwongeecs: keep the private var `clearTextIcon` like before and use it here. Maybe rename it to something like `clearTextButton`.

berkeley-mobile/Home/Search/SearchBarView.swift
```diff
@@ -65,31 +70,56 @@ struct SearchBarView: View {
             isFocused = newValue
             onSearchBarTap?(newValue)
         }
+        .sheet(isPresented: $isPresentingFeedbackForm) {
+            Group {
+                if let config = feedbackFormConfig {
+                    FeedbackFormView(viewModel: feedbackFormViewModel, config: config)
```
justinwongeecs: Because of my recent PR of using Factory for our dependency injection system, we don't need to pass view models into the views as instead we are resolving the view models right within the views themselves.

So you can remove them the `viewModel` parameter and its respective argument.

berkeley-mobile/Home/Search/SearchBarView.swift
```diff
@@ -14,7 +14,10 @@ struct SearchBarView: View {
     @InjectedObject(\.searchViewModel) private var viewModel
 
     @FocusState private var isFocused: Bool
-
+    @State private var isPresentingFeedbackForm = false
+    @State private var feedbackFormViewModel = FeedbackFormViewModel()
```
justinwongeecs: Replace with `@InjectedObservable(\.feedbackFormViewModel) private var feedbackFromViewModel` after line 14

berkeley-mobile/Home/Search/SearchBarView.swift
```diff
@@ -65,31 +70,56 @@ struct SearchBarView: View {
...
+                if feedbackFormConfig != nil {
+                    isPresentingFeedbackForm = true
+                }
+            }
         }) {
-            Image(systemName: "xmark")
+            Image(systemName: "bubble.left.and.bubble.right")
                 .foregroundStyle(Color(BMColor.searchBarIconColor))
                 .fontWeight(.semibold)
                 .frame(width: 30, alignment: .center)
```
justinwongeecs: The search/back SF symbol is `20` points, so I think we should change it back to `20` to make both of the symbols the same size. Because the feedback bubble symbol seems a bit large.

PR #483, opened 2025-11-22
Ananya/user location guide

### LINKED ISSUE #462: Show User Location For Each Guide Place
Like what we do in the list of dining halls of showing the distance of the current user to that dining hall, we should show the distance for Guide Place in `GuideDetailView`.

ananyadua27:
This PR adds the missing distance label to each Guide Place inside the GuideDetailView.
The goal is to match the behavior already used in Dining, Fitness, and Study views, where each location shows how far it is from the current user.

Previously, the distance-to-user display was implemented inline in HomeSectionListRowView. We now extract that implementation into a shared DistanceLabelView component and use it in both the Home drawer and Guide detail views.

**What Was Changed**
- Created a new reusable DistanceLabelView component with the walking figure icon and distance formatting
- Refactored HomeSectionListRowView to use the new DistanceLabelView component instead of inline implementation
- Added distance label to each Guide Place inside GuideDetailRowHeaderView using the same component
- Distance display now consistent across Home drawer (Dining, Fitness, Study) and Guide detail views

**Testing**
- Built and tested on iOS Simulator.
- Confirmed distance displays correctly once user location is available.
- Verified web, phone, and map buttons still function normally.
- Verified no layout shifts or broken overlays on the header image.

Closes #462 

Screenshot 2025-11-24 at 8 18 11 PM

### DIFF
--- a/berkeley-mobile/Home/HomeSectionListRowView.swift
+++ b/berkeley-mobile/Home/HomeSectionListRowView.swift
@@ -7,6 +7,7 @@
 //
 
 import SwiftUI
+
 struct HomeSectionListRowView: View {
     var rowItem: SearchItem & HasLocation & HasImage
     


berkeley-mobile/Common/FilterView/BMSortEngine.swift
justinwongeecs: I don't think this file and all the files under `Common/FilterView` should be in this PR

berkeley-mobile/Home/BMHomeSectionListView.swift
justinwongeecs: I don't think the changes in this file should be part of this PR?

ananyadua27: I was working on two issues almost parallelly. These files are a part of this PR: https://github.com/asuc-octo/berkeley-mobile-ios/pull/482. Sorry for the confusion!

berkeley-mobile/Common/FilterView/BMSortEngine.swift
ananyadua27: I was working on two issues almost parallelly. These files are a part of this PR: https://github.com/asuc-octo/berkeley-mobile-ios/pull/482. Sorry for the confusion!

berkeley-mobile/Home/HomeSectionListRowView.swift
```diff
@@ -44,15 +44,7 @@ struct HomeSectionListRowView: View {
...
-        HStack {
-            Image(systemName: "figure.walk")
-                .foregroundStyle(Color(BMColor.blackText))
-                .font(.system(size: 12))
-            
-            Text("\(rowItem.distanceToUser ?? 0.0, specifier: "%.1f") mi")
-                .foregroundStyle(Color(BMColor.blackText))
-                .font(Font(BMFont.light(12)))
-        }
+        DistanceLabelView(distance: rowItem.distanceToUser)
```
justinwongeecs: We can remove the `distanceLabelView` private variable all together because it just contains `DistanceLabelView`. 

You can use `DistanceLabelView` directly in the view's body.

PR #557, opened 2026-03-31
Convert ResourcesViewModel To Use Observation

### LINKED ISSUE #549: Convert `ResourcesViewModel` To Use `Observation`
Currently `ResourcesViewModel` uses the `ObservableObject` paradigm. We should convert the view model to use the `Observation` framework introduced since iOS 17.

megangadiparthy:
Currently ResourcesViewModel uses the ObservableObject paradigm. We should convert the view model to use the Observation framework introduced since iOS 17.

### DIFF
--- a/GoogleService-Info.plist
+++ b/GoogleService-Info.plist
@@ -1,0 +1,38 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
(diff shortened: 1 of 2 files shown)


GoogleService-Info.plist
justinwongeecs, line 1: We shouldn't check this file into git. 

You can remove this file from being tracked by git by I believe via the `git restore [PATH TO FILE]` command.

Secrets.swift
justinwongeecs, line 1: We shouldn't check in this file as well.

justinwongeecs (changes requested):
We shouldn't check in `GoogleService-Info.plist` and `Secrets.swift`. Also I don't think I see any changes checked in for `ResourcesViewModel.swift`?

Issue #535, opened 2026-03-08
Fix `Comparison with '.nan' using '!=' is always true, use '!lat.isNaN' to check if 'lat' is a number` Warning


justinwongeecs:
In `HasLocation.swift` fix the warning:

```
Comparison with '.nan' using '!=' is always true, use '!lat.isNaN' to check if 'lat' is a number
```

PR #559, opened 2026-04-02
Fix NaN comparison warning (#535)

juhithaclemens:
Replaced != Double.nan checks with !value.isNaN in HasLocation and MapViewController.


justinwongeecs (approved):
lgtm!

PR #563, opened 2026-04-07
Remove oldMapView in SafetyMapView

### LINKED ISSUE #532: Remove oldMapView in SafetyMapView
The minimum deployment target version is 18.0 which we means for new updates we only support devices running 18.0 and above. 

In `SafetyMapView.swift` we are conditionalizing on the operating system version so that for devices running iOS 16.0 and older we show `oldMapView` and newer devices we show `SafetyNewMapView`. 

We should always show `SafetyNewMapView`.

kmiyberk28-afk:
The minimum deployment target version is 18.0 which we means for new updates we only support devices running 18.0 and above.

In SafetyMapView.swift we are conditionalizing on the operating system version so that for devices running iOS 16.0 and older we show oldMapView and newer devices we show SafetyNewMapView.

We should always show SafetyNewMapView.

### DIFF
--- a/berkeley-mobile/Safety/SafetyMapView.swift
+++ b/berkeley-mobile/Safety/SafetyMapView.swift
@@ -18,14 +18,12 @@ struct SafetyMapView: View {
     var isPresentingDetailView: Bool
     
     var body: some View {
-        if #available(iOS 17.0, *) {
-            SafetyNewMapView(selectedSafetyLog: $selectedSafetyLog,
-                             isShowingLegend: $isShowingLegend,
-                             drawerViewState: $drawerViewState,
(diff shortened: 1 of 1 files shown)


berkeley-mobile/Safety/SafetyMapView.swift
```diff
@@ -54,6 +44,8 @@ struct SafetyNewMapView: View {
     
     private let bounds = MapCameraBounds(centerCoordinateBounds: BMConstants.mapBoundsRegion, maximumDistance: BMConstants.mapMaxZoomDistance)
     
+    //PR Check
+
```
justinwongeecs: Remove

justinwongeecs (changes requested):

PR #553, opened 2026-03-23
Kojiro/sort items by user location

KojiroTsugaru:
Sorts items in the home drawer section list by proximity to the user's current location.
 
Changes:
- Added `sortedItems` computed property that sorts items by `distanceToUser` in ascending order
- `pinnedItems` and `nonPinnedItems` now reference `sortedItems` instead of `items` directly

Issue: https://github.com/asuc-octo/berkeley-mobile-ios/issues/436

PR #564, opened 2026-04-07
sort home drawer section item by distance to user

KojiroTsugaru:
Recreated #553 from non-forked repo.

Sorts items in the home drawer section list by proximity to the user's current location.
 
Changes:
- Added `sortedItems` computed property that sorts items by `distanceToUser` in ascending order
- `pinnedItems` and `nonPinnedItems` now reference `sortedItems` instead of `items` directly

Issue: https://github.com/asuc-octo/berkeley-mobile-ios/issues/436


justinwongeecs (approved):
lgtm!

PR #571, opened 2026-04-27
Add search bar UI to Resources tab

kmiyberk28-afk:
The resources tab contains different links grouped into subgroups within three categories: "Wellbeing", "Campus Life", and "Academics".

To help users better find links, we would like to add a search bar that not only show results based on matching the exact text but also intelligently surface results based on a natural language search query.

For example, "What activities can I do around campus?" and it can suggest the "Events and Attractions" subgroup under "Campus Life". This cannot be done through regular string text matching and requires natural language understanding.

We will first start with a basic search query functionality based on matching text, then implement natural language results through an on-device LLM (Apple's Foundational Models). feature #528


Devansh-Ojha:
LGTM!

Devansh-Ojha (approved):
LGTM!

